PR #1, opened 2015-04-08
Add support for object creation through js.new directive.

sbriais:


alainfrisch:
Great, thanks!

Can you please add the following?
- Adapt @@js.new so as to force that the implicit name starts with new_ (and drop this prefix), as for property setter.
- A description of the automatic heuristic in README.md.
- An example in main.html/main.ml.

PR #4, opened 2015-04-09
Enum default

sbriais:
Add support for default cases in enumerations (int or string).

type t = Foo | Bar | Other of string [@js.default]
val f : [`foo |`bar | `other of string [@js.default]] -> ...


alainfrisch:
Thanks!

PR #37, opened 2015-05-28
Partial support for GADT used as enum types.

sbriais:
This provides support for such construction:

type _ event = 
| MouseOver: mouse_event event [@js "mouseover"]
| KeyDown: keyboard_event event [@js "keydown"]

by generating the function 

val event_to_js: 'a event -> Ojs.t

but not the converse function.

### DIFF
--- a/gen_js_api.ml
+++ b/gen_js_api.ml
@@ -908,8 +908,7 @@ and js2ml_unit ty_res res =
 and gen_typ = function
   | Name (s, tyl) ->
       Typ.constr (mknoloc (Longident.parse s)) (List.map gen_typ tyl)
-  | Js ->
-      Typ.constr (mknoloc (Longident.parse "Ojs.t")) []
+  | Js -> ojs_typ
   | Unit _ ->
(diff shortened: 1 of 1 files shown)


alainfrisch:
Currently undocumented, and this might go away.

ghost:
@alainfrisch Is this implemented? I would love to be able to use GADTs for enums. I tried it on the following example but it complains about not being able to parse the type in `t_to_js`:

``` ocaml
module Foo : sig
  type _ t =
    | Bar : int t [@js 0]
    | Baz : string t [@js "baz"]
    [@@js.enum]
  val t_to_js : 'a t -> Ojs.t
end = [%js]
```

sbriais:
Hello,
I believe a workaround is:

module Foo : sig
  type _ t =
    | Bar : int t [@js 0]
    | Baz : string t [@js "baz"]
    [@@js.enum]
  [@@@js.stop]
  val t_to_js : 'a t -> Ojs.t
  [@@@js.start]
end = [%js]

ghost:
@sbriais that seems to work, thanks!

PR #42, opened 2015-06-04
Verbatim

sbriais:
Add support for js.verbatim attribute.

### DIFF
--- a/NAMING.md
+++ b/NAMING.md
@@ -26,7 +26,7 @@ type myType = { x_coord : int; y_coord : int [@js "Y"]}
 is mapped to a JS record with two fields named "xCoord" and "Y" whereas
 
 ```ocaml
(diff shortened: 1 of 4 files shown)


alainfrisch:
- I'd suggest to rename verbatim to something more explicit (e.g. "verbatim_names").  One could imagine for instance that "verbatim" applied e.g. to a type/exception/.. definition would copy the definition verbatim in the implementation.  (See e.g. the existing "parse_sig_verbatim" in the code.)
- Instead of simply collecting any "surrounding" attributes, wouldn't it be better to parse the ones supposed to "flow down" (currently, only this "verbatim_names" attributes, no?) into a structured record?

sbriais:
The attributes are not parsed too early in order for the "spurious attribute" error to work.
Added handling of a js.scope to ease access to JS global values fields.

PR #44, opened 2015-06-15
Default case in sums.

sbriais:


mlasson:
This PR has been integrated at some point.

Issue #58, opened 2016-06-15
Problem using sum types


o-marshmallow:
Hi !

I have been using gen_js_api for days and it works like a charm ! However, I am encountering a problem with sum types. Indeed, I created a file try.mli containing :
`type t =
  | Foo [@js "foo"]
  | Bar [@js 42]
  | Baz
    [@@js.enum]` 
(It's the example given in the documentation)

When I compile it (with all the commands given), the command
 `ocamlfind ocamlc -c -package gen_js_api.ppx try.ml` returns the error :
`File "try.ml", line 4, characters 10-12:
Error: Spurious js.* attribute`

In fact, the `.ml` file still contains the annotations `[@js "foo"]`, `[@js 42]` and `[@@js.enum]`. If I remove then from the .ml file, then it compiles and works well. So is it normal to still have the annotations in the `.ml` file ?

Thanks

sbriais:
Hello, thanks for your report.
A simple workaround is to remove the call to gen_js_api.ppx when compiling with ocamlc. 
We will fix that in the near future.
Cheers
Sébastien

sbriais:
Just pushed a fix onto master.

o-marshmallow:
It's working now, 
Thank you !

sbriais:
Nice!

PR #59, opened 2016-07-21
Corrected new constructor names

o-marshmallow:
The names of the constructors changed.
Moreover, is it possible for you to update the gen_js_api package in the official opam repo? (Issue #58)


alainfrisch:
Hmm, actually this was already done in bf015050a7914c473b853a2aa9860a16e91ff3be, which explains that the released version compiled with OCaml 4.03.  But it was then reverted by mistake in 7b0caf37d4ad1003f977cc217785429b576d4ed2.

Thanks!

Issue #66, opened 2016-11-14
Bindings with requires


Aaylor:
Considering a javascript library containing a module A with only one static
function called **foo** taking a _string_ and returning an _int_. This library
has to be required before usage, but I did not found anything about _require_ in
the documentation.

The first implementation I made was:
```ml
module A : sig
  type t
  val require_a : unit -> t
    [@@js.custom let require_a () = Helpers.require "A" ]
  val foo : t -> string -> int
end
```

To call the function **foo** in ocaml code, I have to write:
```ml
let module_a = A.require_a () in
let i = A.foo module_a "bar" in
...
```

But I'd like to get rid of the require line. This require would be done lazily into the
module, allowing the user to write directly `let i = A.foo "bar"` to call the
function.

My first thought was to do something like:
```ml
module A : sig
  [@@@js.implem
    let mod_a = ref None
    let get_a () =
      match !mod_a with
      | Some m -> m
      | None ->
        let m = Helpers.require "A" in
        mode_a := Some m;
        m
  ]
  val foo : string -> int
  [@@js.custom
    let foo str =
      let m = get_a () in
      let r = Ojs.call m "foo" [|(Ojs.string_to_js str)|] in
      Ojs.int_of_js r ]
end
```

But it becomes very annoying to write this when the javascript library has many
methods. What would be the best way to do this kind of things ? Does it match
the _philosophy_ of javascript bindings ?

alainfrisch:
First, even with the manual encoding, you could use gen_js_api to produce a binding for:

``` ocaml
  type a_module = Ojs.t
  val foo_internal: a_module -> string -> int
```

and then simply define the user-visible `foo` in term of this `foo_internal`:

``` ocaml
  let a = lazy (...)
  let foo x = foo_internal (Lazy.force a) x
```

Now, assuming that a Javascript global variable `A_module` is bound to the result of `require("A")` (either with manual JS code, or through gen_js_api), you can also directly expose:

``` ocaml
  val foo: string -> int [@@js.global "A_module.foo"]
```

One could also design some built-in gen_js_api support for this scheme, perhaps bound to OCaml modules:

``` ocaml
module A : sig
  val foo: string -> int
end [@@js.require]
```

within a module marked with the `js.require` attribute, all functions would implicitly be assumed to be on the object resulting from a call to `require` (by default, with a name derived from the module binding, with a way to override that).

Aaylor:
Thanks for your answer !
[...]

PR #68, opened 2016-12-02
Add require attribute.

Aaylor:
All information about how works the require attribute are in the document REQUIRE.md; it contains three kind of usage, and the meaning of each automatic rules when under the require context.

Related to the discussion #66.

### DIFF
--- a/REQUIRE.md
+++ b/REQUIRE.md
@@ -1,0 +1,124 @@
+Value bindings using required object in gen_js_api
+=========================================================
+
+It happens sometime that, when writing bindings of js library, there
+is a need to require some modules before using it.
+
+```js
+const myModule = require('myModule')
+/* ... */
+```
+
+There is a support to automatically write bindings and to use the
+required object instead of using the global object.
+
+The _require_ attribute
+--------------------------
+
+### Global ###
+
+The attribute `[@@@js.require "myModule"]` indicate to subsequents
+declarations to use the require module **myModule** during the
+generation phase. This global attribute requires the module name, it
+cannot be inferred.
+
+### Local to a module ###
+
+It's also permitted to attach the require attribute to modules.
+
+```ocaml
+module MyModule : sig
+  ...
+end [@@js.require]
+```
+
+By default, the name of the module is used as the name of the js
+module to require.
+
+### Associated to a variable ###
+
+To use bindings rules describe in [Value section](VALUES.md), it's
+possible to put the annotation to a simple variable.
+
+```ocaml
+type t = private Ojs.t
+
+val module_t : t Lazy.t [@@js.require "myModule"]
+
+val get_x : t -> int
+```
+
+###### Nested requires ######
+
+Actually, it's not possible to nest requires.
+
+How is it translated ?
+-------------------------
+
+First, it creates a definition which call lazily the requires
+function:
+
+```ocaml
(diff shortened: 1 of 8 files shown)


REQUIRE.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Associated to a variable ###
+
+To use bindings rules describe in [Value section](VALUES.md), it's
+possible to put the annotation to a simple variable.
+
+```ocaml
+type t = private Ojs.t
+
+val module_t : t Lazy.t [@@js.require "myModule"]
```
alainfrisch, line 46: Should this work as well with a non-lazy binding?

Aaylor: It will not work with non-lazy bindings yet. I force the use of lazyness to avoid requiring javascript module when they're not necessary.

But i can support both version (one lazy, and one non-lazy) if needed.

dannywillems:
@Aaylor I didn't read the PR, but is `require` returning functions (like nightmare, see [this partial binding to Node](https://github.com/besport/ocaml-node) with gen_js_api) supported?

Aaylor:
Actually it's not supported, but it's just a simple case to add; but it will need a relaxed type for the `require` function (which is actually `string -> Ojs.t`).

A first solution would be to declare the `require` function as unsafe, and give it the type `string -> 'a`
The second solution should be to split in two part: a safe `require` and the unsafe one; and detect when the unsafe one should be called. 

With one of the two solution above, we would be able to write:
```ocaml
val my_fun_module : (t1 -> ... -> tn) Lazy.t [@@js.require "myModule"]
```

Aaylor:
@dannywillems Well infact, with this version, it's possible to have function but it's not convenient. Requiring a function could be written as

```ocaml
val my_fun_module : Ojs.t Lazy.t [@@js.require "myModule"]
```
and then used as following:
```ocaml
let _ =
  let my_fun_m = Lazy.force my_fun_module in
  let my_result_ojs = Ojs.apply my_fun_m [| (* ... *) |] in
  (* ... *)
```

I'm not sure that this is a philosophy of gen_js_api to give an `unsafe` function as I proposed earlier, @alainfrisch ? Is this a better thing to have an `unsafe_require` with type `string -> 'a`, or to avoid it ?

src/gen_js_api.ml
```diff
@@ -444,55 +456,91 @@ let parse_attr ~global_attrs ?ty (s, loc, auto) (k, v) =
...
+    | None -> None
+    | Some (k, v) ->
+        if require_ctx then error k.loc (Not_supported_here "nested requires")
+        else
+          match v with
+          | PStr [] -> Some default
+          | _ -> Some (id_of_expr (expr_of_payload k.loc v))
+  in
+  require, require_ctx || require <> None
+
```
sbriais, line 471: I am not sure it's a good idea that this function returns a pair, since the second projection can be simply obtained from the first component and the require_ctx argument. My guess is that it would simplify the code a bit to move the corresponding logic at call sites (when needed).

sbriais:
Could you please provide a full example (if I am not mistaken, the js part is missing)?
Ideally, a concrete example would even be better...

Lupus:
Sad to see this abandoned. Just stumbled across this problem when trying to bind some nodejs to jsoo with gen_js_api myself and hand to resort to defining globals and using `@@js.global "...."` annotation.

PR #76, opened 2017-11-20
adapt to OCaml 4.06 changes to AST

damiendoligez:
There was a small change to the AST in OCaml 4.06. This changes the `gen_js_api` source accordingly.


sbriais:
Many thanks!

PR #83, opened 2019-06-22
Port to dune & ocaml-migrate-parsetree

jchavarri:
This is an attempt (in a very early stage) to port gen_js_api to use dune + ocaml-migrate-parsetree, so it can be more easily consumed by other projects and benefit from the advantages of omp drivers.

The ppx is compiling as a driver already, but the PR is far from done. There are several tasks remaining, which I'm not sure how to solve, but sharing them here in case anyone can offer any advice or guidance:

- Fix `Unimplemented Javascript primitive caml_pure_js_expr!` errors when consuming the ppx from other projects. It seems that `ojs.ml` makes use of externals that are defined in js_of_ocaml, so there must be missing `.js` files or libraries that belong to the js_of_ocaml compiler that would need to be linked together with the ppx lib. I'm not sure how this can be solved.
- Move the standalone binary logic (command line arguments handling etc) to `/bin`
- Update examples so they build with dune as well
- Update documentation to reflect changes
- Release tasks (version upgrade etc)


jchavarri:
Some notes after discussing about this with @rgrinberg on Discord:

```
(library
  (name gen_js_api)
  (public_name gen_js_api.ppx)
  (libraries js_of_ocaml ocaml-migrate-parsetree)
  (c_names ojs_runtime_stubs)
  (js_of_ocaml
    (javascript_files ojs_runtime.js))
  (kind ppx_rewriter))
```

> This seems wrong to me. why would a rewriter ever include jsoo runtime files?
A rewriter can only be ran in native/bytecode. The ppx stack does not work in JS. At least not yet

> you should not use js_of_ocaml when defining your process:
this cannot work b/c this introduces runtime primtives that only work in the browser

> you shouldn't ojs_runtime.js as a js runtime file for your rewriter:
Rewriters cannot use or run any JS code. the ojs_runtime.js should be moved to its own library (gen_js_api.runtime sounds like a good name to me) and added as a ppx_runtime_libraries for your rewriter

> > but isn't the rewriter responsible of declaring it as dep then? it seems fragile otherwise

> Indeed. ppx_runtime_libraries is the mechanism for this.

nojb:
Hello, this project is used internally at LexiFi and we would rather not depend on `ocaml-migrate-parsetree` for the time being. Instead of doing as in your PR, I would suggest simply separating the "driver" code from `gen_js_api.ml` (the part that you commented out) into a different module and expose the ppx mapper as a library so that integration with OMP can be done by an external project.

@sbriais and @alainfrisch may have other suggestions.

hhugo:
- Is there any objection against switching to dune ?

- What is the rational for not depending on ocaml-migrate-parsetree ? The reason I ask is that I was thinking about switching to ppxlib (which relies on ocaml-migrate-parsetree). Ppxlib allows some nice checks, like checking for unused attributes (because of typos).

nojb:
> * Is there any objection against switching to dune ?
> * What is the rational for not depending on ocaml-migrate-parsetree ? The reason I ask is that I was thinking about switching to ppxlib (which relies on ocaml-migrate-parsetree). Ppxlib allows some nice checks, like checking for unused attributes (because of typos).

Hi @hhugo, thanks for your questions. The main issue is that internally we build on Windows and use `omake`. This means no `opam` and no `ocamlfind` (for different reasons) which means additional third-party dependencies need to have a good justification :)

Having said that we are trying to migrate to `dune` which would make all this much easier.

Going back to your question, since `gen_js_api` is a rather small project, do you think it would be possible to arrange things so that a "ppxlib" driver can be written separately while maintaining the possibility of compiling without it, like it is done today? (I have never used ppxlib myself) Similarly, I think adding some `dune` files is OK as long as we can continue to build without it.

nojb:
See #86

PR #86, opened 2019-07-31
introduce a new lib: gen_js_api_ppx

hhugo:
This change aims at making future integration with ppxlib & ocaml-migrate-parsetree easier.


nojb (approved):
Looks good. Do you mind moving the `ppx` directory to the toplevel and maybe renaming the `src` directory `lib` or `runtime` to keep a flatter directory structure? (am not sure what is the canonical layout for ppx's).

Also, if you switch to dune then we don't need to keep the Makefiles anymore as we don't use them internally.

Thanks!

hhugo:
I'm going to flatten the directory structure together with the switch to dune.
Please wait a few days

nojb:
> I'm going to flatten the directory structure together with the switch to dune.
> Please wait a few days

OK, are you going to do that as part of this PR or shall we merge this one first?

hhugo:
If you're happy reviewing a single bigger PR. We could do everything in  
https://github.com/LexiFi/gen_js_api/pull/87

nojb:
> If you're happy reviewing a single bigger PR. We could do everything in
> #87

Sure.

PR #87, opened 2019-08-01
Switch to dune

hhugo:
Replace https://github.com/LexiFi/gen_js_api/pull/83
- [x] switch existing libraries to dune
- [x] introduce gen_js_api_ppx_lib and gen_js_api_ppx_driver
- [x] switch examples to dune
- [x] delete old makefiles 
- [x] add package description
- [x] verify the list of installed files
- [x] verify generated META
- [x] update doc


hhugo:
cc @jchavarri

hhugo:
@nojb, should be ready for review.

.gitignore
```diff
@@ -10,4 +10,7 @@
 *.exe
 examples/*/*.js
 examples/test/test_bindings.ml
-src/gen_js_api
+src/ppx/bin/gen_js_api
```
nojb: I guess all the above are no longer needed ?

lib/dune
```diff
@@ -0,0 +1,10 @@
+(library
+  (name gen_js_api)
+  (public_name gen_js_api)
+  (synopsis "Runtime support for gen_js_api")
+  (libraries)
```
hhugo, line 5: Adding a dependencies on js_of_ocaml would probably be more accurate as gen_js_api depends on stubs provided by js_of_ocaml. @nojb, any opinion ?

nojb (approved):
Looks good. Have you checked if the `ocamlfind` instructions in `INSTALL_AND_USE.md` continue to be valid? Thanks!

lib/dune
nojb: Does adding this dependency have any observable difference for users of the library, apart from requiring them to install `js_of_ocaml`?

hhugo:
> Looks good. Have you checked if the ocamlfind instructions in INSTALL_AND_USE.md continue to be valid? Thanks!

Still works.

lib/dune
hhugo: When compiling js from a bytecode executable, the linker will hopefully not link any of the js_of_ocaml unit as they won't be referenced anywhere (but with -linkall).

When doing separate compilation, the js_of_ocaml units will be included.

nojb:
This is ready for merging on my side. Anything left to do on your side?

hhugo:
I've got one more commit that enable the default set of warnings https://github.com/hhugo/gen_js_api/commit/31807483d849af84b1cc26b744794a0dbad23062. Will open another PR once this one is merged.

lib/dune
nojb: > When doing separate compilation, the js_of_ocaml units will be included.

Will this be the case if "implicit transitive deps" in dune is false ?

hhugo: I would assume so.

nojb: Ok, I don't feel strongly about this, add the dependency if you think so. Otherwise we can merge directly.

hhugo: An alternative would be to generate dummy stubs similar to what jsoo does 
```
(rule
 (targets js_of_ocaml_stubs.c)
 (deps (glob_files *.ml))
 (mode promote)
 (action (with-stdout-to %{targets}
          (run ./gen_stubs/gen_stubs.exe %{deps}))))
```

nojb: Indeed; PRs always welcome :)
Another approach would be to have `dune` pass `-no-check-prims` when an executable is compiled only in js mode.
Anyway, I think the PR is ready to be merged; improving this point can be done in a separate PR.
Shall we merge?

hhugo: Please do

PR #89, opened 2019-08-02
Update the opam file

hhugo:
Note that with theses changes, one should be able to use `dune-release` to release new versions.

### DIFF
--- a/gen_js_api.opam
+++ b/gen_js_api.opam
@@ -12,9 +12,10 @@ license: "MIT"
 dev-repo: "git+https://github.com/LexiFi/gen_js_api.git"
 build: [["dune" "build" "-p" name "-j" jobs]]
 depends: [
-  "ocaml" {>= "4.06.0" & < "4.09.0"}
+  "ocaml" {>= "4.06.0"}
   "dune" {build & >= "1.11"}
   "js_of_ocaml"  {>= "3.1.0"}
+  "ocaml-migrate-parsetree" {>= "1.4.0"}
 ]
 synopsis: "Easy OCaml bindings for Javascript libraries"
 description: """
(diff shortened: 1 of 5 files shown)


gen_js_api.opam
```diff
@@ -9,10 +9,18 @@ authors: [
 homepage: "https://github.com/LexiFi/gen_js_api"
 bug-reports: "https://github.com/LexiFi/gen_js_api/issues"
 license: "MIT"
-dev-repo: "https://github.com/LexiFi/gen_js_api.git"
+dev-repo: "git+https://github.com/LexiFi/gen_js_api.git"
 build: [["dune" "build" "-p" name "-j" jobs]]
 depends: [
+  "ocaml" {>= "4.06.0" & < "4.09.0"}
```
nojb: Why is the `< 4.09.0` contraint needed ?

hhugo: I migth have taken this from the opam-repository. Might not be needed anymore

nojb: Also, don't we need `ocaml-migrate-parsetree` as well?

hhugo: Indeed. I won't be able to do any change in the next few days. Feel free to push to the branch directly if you want to move the change forward.

hhugo: fixed

nojb (approved):
LGTM. Are you planning to add anything else to this PR or is it ready to merge? Thanks!

hhugo:
Ready to merge

hhugo:
I've just realized that it needs at least OCaml 4.08.0. I've updated the constraint.

@nojb, Do you have any ETA for switching to dune at Lexify ? Adding a dependency on 
ocaml-migrate-parsetree would allow gen_js_api to work on any version.

nojb:
> @nojb, Do you have any ETA for switching to dune at Lexify ?

Unfortunately, no. I am working on it as fast as possible, but other priorities come up regularly which makes us go slower than we would want. Also, sometimes we come across the need to extend `dune` to handle our codebase which also takes time.

nojb:
Thanks!

PR #90, opened 2019-08-17
Integrate with ppxlib

hhugo:
Following suggestion in https://github.com/LexiFi/gen_js_api/pull/83#issuecomment-516743718

### DIFF
--- a/ppx-lib/gen_js_api_ppx.ml
+++ b/ppx-lib/gen_js_api_ppx.ml
@@ -1617,8 +1617,14 @@ let mapper =
       check_loc_mapper.Ast_mapper.structure
         check_loc_mapper
         (mapper.Ast_mapper.structure mapper str));
-    signature = (fun this s ->
-      (* run [str_of_sg] to mark `js.***` attributes as used. *)
-      let (_ : structure) = str_of_sg ~global_attrs:[] s in
-      this.signature mapper s)
   }
+
+let mark_attributes_as_used mapper =
+  (* mark `js.***` attributes as used in mli. *)
+  let attribute : _ -> attribute -> _ =
+    fun this ({attr_name = {txt; _}; _} as attr) ->
+      if is_js_attribute txt then
+        ignore (filter_attr txt attr : bool);
+      mapper.Ast_mapper.attribute this attr
+  in
+  { mapper with attribute }
--- a/ppx-lib/gen_js_api_ppx.ml
+++ b/ppx-lib/gen_js_api_ppx.ml
@@ -40,19 +40,27 @@ type error =
 
 exception Error of Location.t * error
 
+let check_attribute = ref true
+let mark_as_handled_manually = ref (fun (_ : Parsetree.attribute) -> ())
 let used_attributes_tbl = Hashtbl.create 16
-let register_loc loc = Hashtbl.replace used_attributes_tbl loc ()
+
+let register_loc attr =
+  !mark_as_handled_manually attr;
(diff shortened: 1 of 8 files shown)


hhugo:
cc @tyoverby

hhugo:
@nojb, this should be ready for review.

hhugo:
gentle ping

nojb:
@hhugo I'm away at the moment, but will take a look at this on Monday when I'm back. Thanks!

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1509,24 +1530,28 @@ and mapper =
...
-    | Pexp_extension (attr, PTyp ty) when filter_attr_name "js.of" attr ->
+    | Pexp_extension (attr, PTyp ty) when filter_extension "js.of" attr ->
         with_default_loc e.pexp_loc
           (fun () -> ml2js_fun (parse_typ ~global_attrs ty))
     | _ ->
         e
   in
-  {super with module_expr; structure_item; expr; typ}
+  let attribute self a =
+    ignore (filter_attr "js.dummy" a : bool);
```
nojb, line 1546: Can you say a few words about this change ?

hhugo: One of the reason I want to integrate with ppxlib is that ppxlib will complain about any unused attributes. In that context, using an arbitrary attribute for defining `functions that return functions` does not work. I've created a special dummy attribute for that case.

hhugo:
gentle ping

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1580,5 +1605,9 @@ let mapper =
     Ast_mapper.structure = (fun _this str ->
       check_loc_mapper.Ast_mapper.structure
         check_loc_mapper
-        (mapper.Ast_mapper.structure mapper str))
+        (mapper.Ast_mapper.structure mapper str));
+    signature = (fun this s ->
+      (* run [str_of_sg] to mark `js.***` attributes as used. *)
+      let (_ : structure) = str_of_sg ~global_attrs:[] s in
```
nojb: This call to `str_of_sg` breaks the `ppx` mode of `gen_js_api`: for example, every sum type needs to be annotated with a `js.*` attribute, etc. In fact, the `ppx` mode is not supposed to do anything on `.mli` files, see https://github.com/LexiFi/gen_js_api/blob/master/PPX.md for the details.

Perhaps this check can be moved to the `ppxlib` driver ?

nojb (changes requested):
Sorry for the horrible lag, but I finally got around to reviewing this. There is a small non-backwards compatible change in the behaviour of `-ppx` mode with respect to the treatment of `.mli` files.

Other than that, seems OK! Thanks!

ppx-lib/gen_js_api_ppx.ml
hhugo: This is done

hhugo:
Sorry for the long delay. I've rebased the feature and addressed your comment.
Please take a look again.

hhugo:
The CI failure is spurious I believe.

nojb (approved):
LGTM

nojb:
(We are waiting to see if the CI can be fixed)

mlasson:
I think it was a networking issue on the travis side. Restarting the build seems to fix the issue. 
(Hint: you can close/reopen a PR to restart a build request to travis if you don't have admin rights; I restarted the build manually so no need to do that :)).

nojb:
Thanks!

PR #91, opened 2019-08-17
Enable ocamlformat

hhugo:


hhugo:
based on #90

hhugo:
I've rebased the feature on top of master.

hhugo:
This is just a proposal to switch to ocamlformat.
Let me know what you think.

nojb:
Hi @hhugo, sorry for the delay in responding.

Using `ocamlformat` would complicate the synchronization with our internal version of this code. I know about https://github.com/hhugo/merge-fmt and similar tools, but it would be an extra source of friction that we would like to avoid for the time being.

Just FYI, there are some initiatives to switch to `ocamlformat` internally, but nothing concrete yet.

hhugo:
Noted

PR #94, opened 2019-08-30
[typo] correct wrong 'apply_arr' to 'apply'

clembu:
Seemed wrong to define `Ojs.apply_arr` with `Ojs.apply_arr`.


nojb:
Thanks!

PR #101, opened 2020-02-25
Adds travis support

mlasson:
This PR tries to set-up travis and allow to compile gen_js_api with more compiler versions through a better use of ocaml-migrate-parsetree.


Lupus:
Looking at travis configuration seems the answer is yes! Awesome! :+1:

mlasson:
Yes !

PR #103, opened 2020-03-30
Fix locations to pass ppxlib locations invariant checks

hhugo:
Here is explanation for the check (as found in
https://github.com/ocaml-ppx/ppxlib/blob/master/src/location_check.mli)

```
(** Enforces some invariants on AST nodes locations *)

(** {2 What?}
    The invariants are as follow:
    - AST nodes are requested to be well nested wrt. locations
    - the locations of "sibling" AST nodes should not overlap
    {2 Why?}
    This is required for merlin to behave properly.
    Indeed, for almost any query directed at merlin, it will need to inspect the
    context around the user's cursor to give an answer that makes sense. And the
    only input it has to do that is the position of the cursor in the buffer.
    The handling of most queries starts by traversing the AST, using the
    locations of nodes to select the right branch. (1) is necessary to avoid
    discarding subtrees too early, (2) is used to avoid merlin making arbitrary
    choices (if you ask for the type under the cursor, and there seem to be two
    things under the cursor, merlin will need to pick one).
    {2 Guidelines for writing well-behaved ppxes}
    It's obviously not always (indeed rarely) possible to mint new locations
    when manipulating the AST.
    The intended way to deal with locations is this:
    - AST nodes that exist in the source should keep their original location
    - new nodes should be given a "ghost" location (i.e.
    [{ some_loc with loc_ghost = true }]) to indicate that the node doesn't
    exist in the sources.
    Both the new check and merlin will happily traverse these ghost nodes as if
    they didn't exist.
    Note: this comes into play when deciding which nodes are "siblings", for
    instance if your AST is:
    {v
      A (B1(C, D),
         B2(X, Y))
    v}
    but [B2] has a ghost location, then [B1], [X] and [Y] are considered
    siblings.
    Additionally, there is an attribute [[@merlin.hide]] that you can add on
    nodes to tell merlin (and the check) to ignore this node and all of its
    children.
    Some helpers for this are provided in {!Merlin_helpers}.
*)
```

### DIFF
--- a/ppx-lib/gen_js_api_ppx.ml
+++ b/ppx-lib/gen_js_api_ppx.ml
@@ -44,6 +44,14 @@ let check_attribute = ref true
 let mark_as_handled_manually = ref (fun (_ : Parsetree.attribute) -> ())
 let used_attributes_tbl = Hashtbl.create 16
 
+(* [merlin_hide] tells merlin to not look at a node, or at any of its
+   descendants.  *)
+let merlin_hide =
+  { attr_name = { txt = "merlin.hide"; loc = Location.none }
+  ; attr_payload = PStr []
+  ; attr_loc = Location.none
+  }
+
 let register_loc attr =
   !mark_as_handled_manually attr;
   Hashtbl.replace used_attributes_tbl attr.attr_name.loc ()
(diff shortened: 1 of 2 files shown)


hhugo:
cc @trefis

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -44,6 +44,14 @@ let check_attribute = ref true
 let mark_as_handled_manually = ref (fun (_ : Parsetree.attribute) -> ())
 let used_attributes_tbl = Hashtbl.create 16
 
+(* [merlin_hide] tells merlin to not look at a node, or at any of its
+   descendants.  *)
+let merlin_hide =
+  { attr_name = { txt = "merlin.hide"; loc = Location.none }
+  ; attr_payload = PStr []
+  ; attr_loc = Location.none
+  }
```
trefis, line 53: If you are linking with ppxlib, then I believe there is a `Ppxlib.Merlin_helpers` module that you could use, which already defines this attribute.

trefis (approved):
LGTM.

ppx-lib/gen_js_api_ppx.ml
hhugo: Ppxlib is not a dependency here (it's only in the driver)

hhugo:
gentle ping

mlasson:
Thanks for the PR !

PR #110, opened 2020-07-22
Include payload in extension node

### LINKED ISSUE #107: Allow to run ppxlib deriving and gen_js_api ppx together
One currently cannot use ppxlib deriving together with gen_js_api.
In particular, running the code bellow will fail with `Error: Cannot parse signature item`
```ocaml
module M : sig 
  type t = { a : int; b : string } [@@deriving sexp]
  val t_of_js : Ojs.t -> t
  val t_to_js : t -> Ojs.t
end = [%js]
```

The reason is that ppxlib will expand deriving first and then run the gen_js_api mapper, at which point `[@@deriving sexp]` has generated item that gen_js_api doesn't understand (e.g. `include`). 

One could imagine the following steps instead:
1. expand gen_js_api first
```ocaml
module M : sig 
  type t = { a : int; b : string } [@@deriving sexp]
  val t_of_js : Ojs.t -> t
  val t_to_js : t -> Ojs.t
end = struct
  type t = { a : int; b : string } [@@deriving sexp]
  let t_of_js = ....
  let t_to_js = ....
end
```
2. then apply deriving
```ocaml
module M : sig 
  type t = { a : int; b : string }
  include sig
    [@@@ocaml.warning "-333"]
     val t_of_sexp : sexp -> t
     val sexp_of_t : t -> sexp
  end 
  val t_of_js : Ojs.t -> t
  val t_to_js : t -> Ojs.t
end = struct
  type t = { a : int; b : string }
  let t_of_sexp = .....
  let sexp_of_t = .....
  let t_of_js = ....
  let t_to_js = ....
end
```

Talking with @jeremiedimino, there are two approaches to explore:
- making gen_js_api register a `Context_free.Rule.t` instead of whole AST mapper. That might require changing the syntax a little 
- allow to specify some kind of ordering between AST mappers

nojb:
@hhugo 

Fixes #107 (in principle)

### DIFF
--- a/ppx-lib/gen_js_api_ppx.ml
+++ b/ppx-lib/gen_js_api_ppx.ml
@@ -1502,10 +1502,8 @@ and mapper =
   and module_expr self mexp =
     let mexp = super.module_expr self mexp in
     match mexp.pmod_desc with
-    | Pmod_constraint({pmod_desc=Pmod_extension (attr, PStr[]); _}
-                    , ({pmty_desc=Pmty_signature sg; pmty_attributes; pmty_loc = _} as mty))
-         when filter_extension "js" attr ->
-       Mod.constraint_ (Mod.structure ~attrs:[ merlin_hide ] (str_of_sg ~global_attrs:(pmty_attributes) sg)) mty
+    | Pmod_extension ({txt = "js"; _}, PSig sg) ->
+       Mod.constraint_ (Mod.structure ~attrs:[ merlin_hide ] (str_of_sg ~global_attrs:mexp.pmod_attributes sg)) (Mty.signature sg)
     | _ -> mexp
   in
   let structure_item self str =
--- a/PPX.md
+++ b/PPX.md
@@ -24,17 +24,16 @@ Note: the ppx currently does nothing on `.mli` files.
 
 Several forms are supported:
 
- - `[%js]` extension as a module expression, to be used directly under
-   a module-level type constraint.  Examples:
+ - `[%js: <signature>]` extension as a module expression.  Examples:
 
    ````
-     include ([%js] : sig ... end)
+     include [%js: <signature>]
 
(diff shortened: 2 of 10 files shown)


ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1502,10 +1502,8 @@ and mapper =
   and module_expr self mexp =
     let mexp = super.module_expr self mexp in
     match mexp.pmod_desc with
-    | Pmod_constraint({pmod_desc=Pmod_extension (attr, PStr[]); _}
-                    , ({pmty_desc=Pmty_signature sg; pmty_attributes; pmty_loc = _} as mty))
-         when filter_extension "js" attr ->
-       Mod.constraint_ (Mod.structure ~attrs:[ merlin_hide ] (str_of_sg ~global_attrs:(pmty_attributes) sg)) mty
+    | Pmod_extension ({txt = "js"; _}, PSig sg) ->
+       Mod.constraint_ (Mod.structure ~attrs:[ merlin_hide ] (str_of_sg ~global_attrs:[])) (Mty.signature sg)
```
hhugo: Should you set `global_attrs` to `mexp.pmod_attributes`.
Also, you missing an argument to str_of_sg

nojb: Thanks, both are fixed now.

nojb:
@hhugo OK to merge this PR?

hhugo (approved):

nojb:
Thanks!

hhugo:
looks good.
There might be other changes needed. I think we should wait for better ppxlib support to be complete before you cut a release.

nojb:
> There might be other changes needed. I think we should wait for better ppxlib support to be complete before you cut a release.

Yes, of course.

PR #111, opened 2020-07-22
Better ppxlib integration 

### LINKED ISSUE #107: Allow to run ppxlib deriving and gen_js_api ppx together

hhugo:
Better integration with ppxlib. Fix #107
 
As a result, the following syntax changes were made when using the ppxlib driver:
- `val name : ty [@@js...]` in structure is now only supported inside `[%js ]` extension nodes
- `type t [@@js]` in structure not longer expand to `type t = Ojs.t [@@js]`, one need to add the type equality explicitly 
- `[@js replace]`  in structure is now only supported inside `[%js]` extension nodes

### DIFF
--- a/examples/misc/test_jquery.ml
+++ b/examples/misc/test_jquery.ml
@@ -6,8 +6,10 @@
 
 open Jquery
 
+include [%js:
 val alert: string -> unit
   [@@js.global]
+]
 
 let ( !! ) = Jquery.selector
 
--- a/examples/test/main.ml
+++ b/examples/test/main.ml
@@ -40,6 +40,7 @@ let () =
   test_variadic sum;
   test_variadic2 (fun msg xs -> Printf.printf "%s\n%!" msg; sum xs)
 
+include [%js:
 val myArray: int array
     [@@js]
 
@@ -55,6 +56,7 @@ val alert_float: float -> unit
 
 val test_opt_args: (?foo:int -> ?bar:int -> unit-> string) -> unit
   [@@js.global]
+]
 
 let doc = Window.document window
 
@@ -215,6 +217,7 @@ end = struct
     !l
 end
 
+include [%js:
 val int_dict_to_json_string: int Dict.t -> string
   [@@js.global "JSON.stringify"]
 
@@ -226,6 +229,7 @@ val set_x: int -> unit
 
 val get_x: unit -> int
     [@@js.get "x"]
+]
 
 let () =
   print_endline (int_dict_to_json_string ["hello", 1; "world", 2]);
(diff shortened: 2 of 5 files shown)


hhugo:
the CI failure seems unrelated

hhugo:
cc @jeremiedimino

nojb:
> Better integration with ppxlib. Fix #107

Thanks! I will try it out next week.

> As a result, the following syntax changes were made when using the ppxlib driver:
> 
> * `val name : ty [@@js...]` in structure is now only supported inside `[%js ]` extension nodes

This seems reasonable; one can always use the idiom
```
include [%js:
  val name : ty [@@js ...]
]
```
in this case.

> * `type t [@@js]` in structure not longer expand to `type t = Ojs.t [@@js]`, one need to add the type equality explicitly

One alternative is to generate functions
```
external t_of_js : Ojs.t -> t = "%identity"
external t_to_js : t -> Ojs.t = "%identity"
```
in this case.

> * `[@js replace]`  in structure is now only supported inside `[%js]` extension nodes

This seems reasonable (same idiom as above).

hhugo:
> One alternative is to generate functions

It feel like the better alternative is the make the alias explicit, no ? `type t = Ojs.t [@@js]`

nojb:
I played with this a bit; `[@js.replace ...]` does not seem to be working even within a `[%js: ...]` node (it is ignored). Also, if used outside such a node, it should give an error, not be ignored, since otherwise it can silently fail at runtime in the browser which is much worse.

PR #113, opened 2020-10-03
Add support for type variables

### LINKED ISSUE #96: Support type variables on interface files
I find myself doing this quite often in `mli` files that are processed with gen_js_api:

```ocaml
val use_state: (unit -> 'state) -> ('state * ('state -> unit))
[@@js.custom
  val use_state_internal : (unit -> Ojs.t) -> (Ojs.t * (Ojs.t -> unit)) [@@js.global "react.useState"]
  let use_state = Obj.magic use_state_internal
]
```

Binding to functions with type variables is very common, and regular `external` statements support it.

The manual process shown above is verbose and error prone, and I was wondering if it'd be possible for gen_js_api to interpret `'a` as the black box that it is, that doesn't require any conversions (or rather because it can be any type, the only function applicable is identity), and basically automate the "magic" that is needed for those kind of signatures to exist.

jchavarri:
Fixes #96.

I tried to keep the changes to a minimum, please let me know if I'm missing any extra cases.

Added some examples to the test file to show how useful this addition would be to bind to JavaScript APIs that are very dynamic (e.g. `console`, `String`) in a lighter way. 

Side note: I tried to adapt `Console2` module in the tests, but there seems to be a bug when combining `[@@js.global "payload"]` with `[@js.scope "foo"]`, I will try to isolate it and report it in a new issue.

### DIFF
--- a/ppx-lib/gen_js_api_ppx.ml
+++ b/ppx-lib/gen_js_api_ppx.ml
@@ -37,6 +37,7 @@ type error =
   | Spurious_attribute
   | Sum_kind_args
   | Union_without_discriminator
+  | Contravariant_type_parameter of string
 
 exception Error of Location.t * error
 
@@ -167,6 +168,8 @@ let print_error ppf = function
       Format.fprintf ppf "Unknown method to discriminate unions."
   | Union_without_discriminator ->
       Format.fprintf ppf "js.union without way to discriminate values."
+  | Contravariant_type_parameter label ->
+      Format.fprintf ppf "Contravariant type parameter '%s is not allowed." label
 
 let () =
   Location.register_error_of_exn
@@ -412,10 +415,14 @@ let derived_from_type s ty =
   match ty with
   | Arrow {ty_args; ty_vararg = None; unit_arg = false; ty_res = Js} ->
     begin match List.rev ty_args with
-    | [] -> false
     | {lab=Arg; att=_; typ=Name (t, _);} :: _ -> check_suffix ~suffix:"_to_js" s = Some t
+    | _ -> false
     end
-  | Arrow {ty_res = Name (t, _); ty_vararg = None; unit_arg = false } -> check_suffix ~suffix:"_of_js" s = Some t
+  | Arrow {ty_res = Name (t, _); ty_vararg = None; unit_arg = false; ty_args } ->
+    begin match List.rev ty_args with
+    | {lab=Arg; att=_; typ= Js;} :: _ -> check_suffix ~suffix:"_of_js" s = Some t
+    | _ -> false
+    end
   | _ -> false
 
 let auto ~global_attrs s ty =
@@ -795,31 +802,34 @@ let get_variant_kind loc attrs =
         end
   end else error loc (Not_supported_here "Sum types without js.* attribute")
 
-let rec js2ml ty exp =
+let rec js2ml ctx ty exp =
   match ty with
   | Js ->
       exp
   | Name (s, tl) ->
       let s = if builtin_type s then "Ojs." ^ s else s in
-      let args = List.map js2ml_fun tl in
+      let args = List.map (js2ml_fun ctx) tl in
       app (var (s ^ "_of_js")) (nolabel (args @ [exp])) false
   | Arrow {ty_args; ty_vararg; unit_arg; ty_res} ->
-      let formal_args, concrete_args = prepare_args ty_args ty_vararg in
+      let formal_args, concrete_args = prepare_args ctx ty_args ty_vararg in
       let res = ojs_apply_arr exp concrete_args in
-      func formal_args unit_arg (js2ml_unit ty_res res)
+      func formal_args unit_arg (js2ml_unit ctx ty_res res)
   | Unit loc ->
       error loc (Not_supported_here "Unit")
   | Variant {location; global_attrs; attributes; constrs} ->
-      js2ml_of_variant ~variant:true location ~global_attrs attributes constrs exp
+      js2ml_of_variant ctx ~variant:true location ~global_attrs attributes constrs exp
   | Tuple typs ->
       let f x =
-        Exp.tuple (List.mapi (fun i typ -> js2ml typ (ojs "array_get" [x; int i])) typs)
+        Exp.tuple (List.mapi (fun i typ -> js2ml ctx typ (ojs "array_get" [x; int i])) typs)
       in
       let_exp_in exp f
-  | Typ_var _ ->
-      app (var ("Obj.magic")) (nolabel ([exp])) false
+  | Typ_var alpha ->
+      if List.mem alpha ctx then
+        app (var ("__"^alpha ^ "_of_js")) (nolabel ([exp])) false
+      else
+        app (var ("Obj.magic")) (nolabel ([exp])) false
 
-and js2ml_of_variant ~variant loc ~global_attrs attrs constrs exp =
+and js2ml_of_variant ctx ~variant loc ~global_attrs attrs constrs exp =
   let variant_kind = get_variant_kind loc attrs in
   let string_typ = Name ("string", []) in
   let int_typ = Name ("int", []) in
@@ -839,7 +849,7 @@ and js2ml_of_variant ~variant loc ~global_attrs attrs constrs exp =
         | `String s -> int_default, int_cases, string_default, Exp.case (pat_str s) x :: string_cases
         | `Int n -> int_default, Exp.case (pat_int n) x :: int_cases, string_default, string_cases
       in
-      let get_arg key typ = js2ml typ (ojs "get" [exp; str key]) in
+      let get_arg key typ = js2ml ctx typ (ojs "get" [exp; str key]) in
       match arg with
       | Constant -> case (mkval mlconstr None)
       | Unary arg_typ ->
@@ -850,7 +860,7 @@ and js2ml_of_variant ~variant loc ~global_attrs attrs constrs exp =
                 let loc, arg_field = get_string_attribute_default "js.arg" (location, "arg") attributes in
                 check_label loc arg_field;
                 case (mkval mlconstr (Some (get_arg arg_field arg_typ)))
-            | `Union _ -> case (mkval mlconstr (Some (js2ml arg_typ exp)))
+            | `Union _ -> case (mkval mlconstr (Some (js2ml ctx arg_typ exp)))
           in
           let process_default defs cont =
             match get_attribute "js.default" attributes with
@@ -862,7 +872,7 @@ and js2ml_of_variant ~variant loc ~global_attrs attrs constrs exp =
                       let x = fresh() in
                       cont (Exp.case (Pat.var (mknoloc x)) (mkval mlconstr (Some (var x))))
                   | `Sum _ | `Union _ ->
-                      cont (Exp.case (Pat.any ()) (mkval mlconstr (Some (js2ml arg_typ exp))))
+                      cont (Exp.case (Pat.any ()) (mkval mlconstr (Some (js2ml ctx arg_typ exp))))
                 end else error k.loc Multiple_default_case
           in
           begin match variant_kind with
@@ -880,7 +890,7 @@ and js2ml_of_variant ~variant loc ~global_attrs attrs constrs exp =
           | `Sum _ ->
               let loc, args_field = get_string_attribute_default "js.arg" (location, "arg") attributes in
               check_label loc args_field;
-              let get_args key i typ = js2ml typ (ojs "array_get" [ojs "get" [exp; str key]; int i]) in
+              let get_args key i typ = js2ml ctx typ (ojs "array_get" [ojs "get" [exp; str key]; int i]) in
               case (mkval mlconstr (Some (Exp.tuple (List.mapi (get_args args_field) args_typ))))
           | `Union _ -> error location Constructor_in_union
           end
@@ -910,8 +920,8 @@ and js2ml_of_variant ~variant loc ~global_attrs attrs constrs exp =
       | `Union No_discriminator -> error loc Union_without_discriminator
       | `Union (On_field kind) -> ojs "get" [exp; str kind]
     in
-    let int_match = gen_match (js2ml int_typ discriminator) int_default int_cases in
-    let string_match = gen_match (js2ml string_typ discriminator) string_default string_cases in
+    let int_match = gen_match (js2ml ctx int_typ discriminator) int_default int_cases in
+    let string_match = gen_match (js2ml ctx string_typ discriminator) string_default string_cases in
     match int_match, string_match with
     | None, None -> assert false
     | Some int_match, None -> int_match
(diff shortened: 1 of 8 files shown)


ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1219,7 +1227,15 @@ and gen_funs ~global_attrs p =
     match p.ptype_manifest, p.ptype_kind with
     | Some ty, Ptype_abstract ->
         let ty = parse_typ ~global_attrs ty in
-        lazy (js2ml_fun ty), lazy (ml2js_fun ty)
+        (match p.ptype_params, ty with
+        | [], _ ->
+          lazy (js2ml_fun ty), lazy (ml2js_fun ty)
+        | [{ptyp_desc = Ptyp_var _; ptyp_loc = _; ptyp_attributes = _; ptyp_loc_stack = _}, Invariant], Js ->
+          let s = "_" in
```
jchavarri: This covers the case with abstract types with type variable, e.g.

```ocaml
    type 'value t = Ojs.t
```
will generate:
```
    type 'value t = Ojs.t
    let rec (t_of_js : (Ojs.t -> 'value) -> Ojs.t -> 'value t) =
      fun _ -> fun x171 -> x171
    and (t_to_js : ('value -> Ojs.t) -> 'value t -> Ojs.t) =
      fun _ -> fun x170 -> x170
```

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1219,7 +1227,15 @@ and gen_funs ~global_attrs p =
...
+        (match p.ptype_params, ty with
+        | [], _ ->
+          lazy (js2ml_fun ty), lazy (ml2js_fun ty)
+        | [{ptyp_desc = Ptyp_var _; ptyp_loc = _; ptyp_attributes = _; ptyp_loc_stack = _}, Invariant], Js ->
+          let s = "_" in  
+          lazy (fun_ (Nolabel, s) (js2ml_fun ty)),
+          lazy (fun_ (Nolabel, s) (ml2js_fun ty))
+        | _ ->
+          error p.ptype_loc Cannot_parse_type)
     | None, Ptype_abstract ->
```
jchavarri: Side note. I realized this branch actually never runs, because `gen_funs` is called after processing the declarations through `rewrite_typ_decl`:

https://github.com/LexiFi/gen_js_api/blob/4e0f2a344a4d5598619982318ec841b77f923781/ppx-lib/gen_js_api_ppx.ml#L1335

And `rewrite_typ_decl` adds a manifest for this specific case:

https://github.com/LexiFi/gen_js_api/blob/4e0f2a344a4d5598619982318ec841b77f923781/ppx-lib/gen_js_api_ppx.ml#L743

jchavarri (commented):
This is ready for review, I have updated the docs as well to mention the cases cover by the new addition.

Please check `Console3` and `Ref` modules on the examples to see real-world use cases for this new feature, thanks.

mlasson:
Hello ! 

Thanks for this very interesting pull request. 

I think it could be useful indeed but this feature should be handled with care. 

It allows to very quickly use data structures from the javascript world to hold caml values in a "type safe" way. 

```ocaml
module StringMap : sig
    type 'a t = private Ojs.t
    [@@@js.stop]
    val t_to_js: ('a -> Ojs.t) -> 'a t -> Ojs.t
    val t_of_js: (Ojs.t -> 'a) -> Ojs.t -> 'a t
    [@@@js.start]
    val create: unit -> 'a t [@@js.new "Map"]
    val set: 'a t -> string -> 'a -> unit
    val get: 'a t -> string -> 'a option 
end
```

(Note the presence of stop/start is there to get around a small & fixable bug with approach, let's forget about that for now). 

This data structure is great as long as you never use it to interface with a javascript function because a "string StringMap.t" is not holding javascript script strings instead it is holding "ocaml string" (aka. Mlbytes). 

```ocaml
    let open StringMap in
    let table = create () in
    set table "marc" "hello world";
    Js_browser.(Console.log console (t_to_js (fun x -> [%js.of:string] x) table));
 ```
That will print on the console:
```
Map(1) {"marc" => MlBytes}
```

It means that we will not be able to bind a javascript function with signature: 
```ocaml
 val capitalize: string StringMap.t -> unit
```

Do you have examples of functions implemented in the javascript side that handles such polymorphic data structure in a "non-polymorphic" way ?

One example that I have in mind are "Promises":
```ocaml
val fetch: string -> Response.t Promise.t [@@js.global "fetch"]
```
There you want the promise to hold a "javascript response" and not a "caml response". 

I think we should properly document this behavior before merging the PR.  

Would you mind providing examples of use cases ? Do you confirm that the objects stored in your data structures are never handled directly by javascript functions ?

jchavarri:
Hi @mlasson ! Thanks for the feedback, I really appreciate it.

Before answering your comments, I'd like to mention the 2 cases this PR is targeting (or at least, the way I see it):

a) **"Plain" type variables** in arrow functions, like `val foo: 'a ->'a`. This case is simple, but also quite useful already. One example could be the function [`useState`](https://reactjs.org/docs/hooks-reference.html#usestate) from the React.js library, which allows to define stateful values inside React components. The signature is as follows:

```ocaml
val useState : (unit -> 'state) -> 'state * (('state -> 'state) -> unit)
```

The function takes a callback that will set the initial state, and returns a tuple (state, "updater" function). In this case, React.JS does not care which values the user stores in state, so this PR allows to define the function just as written above (see [example in `jsoo-react` bindings](https://github.com/jchavarri/jsoo-react/blob/a631fedbea67acb1f2beca3bec1d1f5f22ddb7f5/lib/react.mli#L245)). 

b) **Abstract types with parameters**, like `'a t`. This case is much more involved 😄 

I bring up these 2 cases because one possibility could be to split the PR in 2 parts, tackling a) and b) in separate steps. Please let me know if this would simplify the review process.

Regarding the feedback from your comments, I understand it's mostly around b). 

> This data structure is great as long as you never use it to interface with a javascript function

Yes, one has to be careful to not use this approach for signatures that expect the "inner" type from `StringMap.t` to be converted to/from JS. I guess this use case could be summarized by all JavaScript "container-like" data structures where the OCaml application will store and read OCaml values "as is".

> Do you have examples of functions implemented in the javascript side that handles such polymorphic data structure in a "non-polymorphic" way ?

I can't think of any examples. All the types I can think of are polymorphic —they are all "containers" of sorts—, and their "non-polymorphism" depends a lot on each application requirements. Promises for example will be used to wrap JavaScript types like `Response.t` in the example you mention 99% of the times. But one can imagine (rare) cases where a promise could wrap OCaml values directly.

One example of such case appears (again) in the js_of_ocaml bindings for React.js. React exposes some feature called ["context"](https://reactjs.org/docs/context.html) that allows to expose data to all the children in a hierarchy of components. Like with `useState`, React does not care about what's stored in context, so one can model it like:

```ocaml
module Context : sig
  type 'props t
end

val createContext : 'a -> 'a Context.t
  [@@js.global "react.createContext"]

val useContext : 'value Context.t -> 'value
  [@@js.global "react.useContext"]
```

[See `react.mli` in repo.](https://github.com/jchavarri/jsoo-react/blob/a631fedbea67acb1f2beca3bec1d1f5f22ddb7f5/lib/react.mli#L284-L319)

> I think we should properly document this behavior before merging the PR.

I agree. I will try to add some examples and explanations that include the details discussed here. Do you think it'd be ok to use the React examples that I mentioned? It's not part of the web standards, but React certainly became a de-facto standard (like jQuery in the past). If there are other examples that are more appropriate, please let me know.

> Do you confirm that the objects stored in your data structures are never handled directly by javascript functions ?

Confirmed 👍

jchavarri:
@mlasson I added some more explanations to reflect the points discussed above in ad6b241. I ended up using your examples because I think they're more generic than the React ones I suggested, and the promise one shows the risks very clearly.

Let me know if more details, or a different way to word this would be better, thanks!

Lupus:
Hm, so how will bindings look like for a `fetch` function? Documentation says what I shouldn't do, but it does not say what I could do instead to bind. May be some example for the same `fetch` function nearby will help?

mlasson:
I made some generalization of what you proposed. 

This PR introduced indeed two somewhat related but distinct features: 
  - The use of "Obj.magic" to derive `'a_to_js` to help to store ocaml values in javascript "containers". 
  - The proper support of "type parameters" (ie. `type 'a t = ...")

What my commit changes: 
  1. I reimlement the "heuristic" to expose of the generated "*_to/of_js" functions to support the generated functions coming the parametrized types. (commit cb049f7) 
  2. I dropped the "Ojs.typvar_of/to_js" in favor of "Obj.magic" (commit cef6c22) (I did that because somewhere in the doc we say that we don't expose "unsafe coercion" in the Ojs.module). It keeps the purple magic under the carpet :).  
  3. Finally, the last two commits implements the "generic support" of parametrized types. It allows to use type variables "soundly". 

Here are some examples of things we could with that: 
```ocaml
(* Suport for more than one type variable *)
module Either : sig
  type ('a, 'b) t
  val t_to_js: ('a -> Ojs.t) -> ('b -> Ojs.t) -> ('a, 'b) t -> Ojs.t
  val t_of_js: (Ojs.t -> 'a) -> (Ojs.t -> 'b) -> Ojs.t -> ('a, 'b) t

  val left: 'a -> ('a, 'b) t [@@js.global "eitherLeft"]
  val right: 'b -> ('a, 'b) t [@@js.global "eitherRight"]
  val destruct: ('a, 'b) t -> left:('a -> 'c) -> right:('b -> 'c) -> 'c [@@js.global "eitherDestruct"]
end
(* Where the corresponding JS code is
      function eitherLeft(anything) {
        return {left: anything};
      }
      function eitherRight(anything) {
        return {right: anything};
      }
      function eitherDestruct(x, left, right) {
        if (x.hasOwnProperty("left")) {
          return left(x.left);
        } else {
          return right(x.right);
        }
      }
*)

 (** "Generic" support of parametrized types: *)
  module Id : sig
    type 'a t = 'a
    val t_to_js: ('a -> Ojs.t) -> 'a t -> Ojs.t
    val t_of_js: (Ojs.t -> 'a) -> Ojs.t -> 'a t
  end

module Pair : sig
    type ('a, 'b) t =
      {
        x: 'a;
        y: 'b
      }
    val t_to_js: ('a -> Ojs.t) -> ('b -> Ojs.t) -> ('a, 'b) t -> Ojs.t
    val t_of_js: (Ojs.t -> 'a) -> (Ojs.t -> 'b) -> Ojs.t -> ('a, 'b) t
  end

  module Swap : sig
    type ('a, 'b) t = ('b, 'a) Pair.t
    val t_to_js: ('a -> Ojs.t) -> ('b -> Ojs.t) -> ('a, 'b) t -> Ojs.t
    val t_of_js: (Ojs.t -> 'a) -> (Ojs.t -> 'b) -> Ojs.t -> ('a, 'b) t
  end

  module Diag : sig
    type 'a t = ('a, 'a) Pair.t
    val t_to_js: ('a -> Ojs.t) -> 'a t -> Ojs.t
    val t_of_js: (Ojs.t -> 'a) -> Ojs.t -> 'a t
  end

 module List: sig 
    type 'a t =
      | Empty
      | Cons of 'a * 'a t
      [@@js.sum]

    val t_to_js: ('a -> Ojs.t) -> 'a t -> Ojs.t
    val t_of_js: (Ojs.t -> 'a) -> Ojs.t -> 'a t
end

  (* Error: Contravariant type parameter !
  module E : sig
    type 'a t = 'a -> int
  end *)
```

mlasson:
@lupus promises are tricky. 

Here's a stub of how you could do it : 

```ocaml
module UntypedPromise : sig

  type t = private Ojs.t
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t

  [@@@js.stop]
    val return: Ojs.t -> t
    val bind: t -> (Ojs.t -> t) -> t
  [@@@js.start]

  [@@@js.implem
    val resolve: Ojs.t -> Ojs.t [@@js.global "Promise.resolve"]
    val then_: Ojs.t -> (Ojs.t -> Ojs.t) -> Ojs.t [@@js.call "then"]

    type wrap = { content: Ojs.t }[@@js]

    let is_promise o =
      resolve o == o

    let wrap o =
      if is_promise o then
        wrap_to_js { content = o }
      else o

    let unwrap o =
      if Ojs.has_property o "content" then
        Ojs.get o "content"
      else
        o

    let return x =
      resolve (wrap x)

    let bind p f =
      then_ p (fun x -> f (unwrap x))

  ]
end

module Promise : sig
  [@@@js.stop]
  type 'a t = UntypedPromise.t
  val t_to_js: ('a -> Ojs.t) -> 'a t -> Ojs.t
  val t_of_js: (Ojs.t -> 'a) -> Ojs.t -> 'a t

  val return: 'a -> 'a t
  val bind: 'a t -> ('a -> 'b t) -> 'b t
  [@@@js.start]
  [@@@js.implem
    type 'a t = UntypedPromise.t
    let return x = UntypedPromise.return (Obj.magic x)
    let bind p f = UntypedPromise.bind p (fun x -> f (Obj.magic x))
    let map f p = bind p (fun x -> return (f x))
    let t_to_js f p = UntypedPromise.t_to_js (map f p)
    let t_of_js f p = map f (UntypedPromise.t_of_js p)
  ]
end

module Response : sig
  type t = private Ojs.t
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t
  val text: t -> string Promise.t  [@@js.call "text"]
end

val fetch: string -> Response.t Promise.t [@@js.global "fetch"]
```

It should give you a "type safe" binding to promises. 
There's a hack to "box" nested promises to get around the crazyness of JS (JS runtimingly flattens promises). 
Here the PR does not  help: the ppx is mostly turned off. 
Note that each time you interface from and to javascript you create a new promise to apply the coercion. 
You can find this two clever tricks in the promise_jsoo package: https://github.com/mnxn/promise_jsoo/blob/90ced28444b6295d6494264953b7d06e5253b5d7/promise.ml#L183

Lupus:
Wow, that looks fantastic! Maybe those bindings should be included into gen_js_api stdlib?

mlasson:
>  Maybe those bindings should be included into gen_js_api stdlib?

Unfortunately there's no such thing !
There are a lot of design choices to provide standard bindings. And as soon as there's more than one user, it starts to be a lot of work to maintain. 

For most things, gen_js makes super easy to incrementally add bindings.
That's why, currently, each project use its own set of bindings and there is no preferred standard library. 
Maybe one day ...

alainfrisch:
Note : https://github.com/LexiFi/ocaml-vdom provides some bindings to the browser's APIs (https://github.com/LexiFi/ocaml-vdom/blob/master/lib/js_browser.mli).

mlasson:
Except if someone objects, I'm about to merge this PR.

Lupus:
I would vote for including the content of https://github.com/LexiFi/gen_js_api/pull/113#issuecomment-707894964 into documentation/examples so that others have stable source for copy-paste into adhoc per-project bindings set :)

But I'm not sure if it's within the scope of this PR, probably not, so I'm all for this good stuff to land to master! :raised_hands:

jchavarri:
> Except if someone objects, I'm about to merge this PR.

Sounds good @mlasson, thanks a lot for your help! There's no way I could have figured this out on my own.

PR #114, opened 2020-10-13
Dependency tweaks

rgrinberg:
* Do not rely on implicit transitive dependencies. Makes it clear what exactly
  depends on omp, ppxlib, and compiler-libs.
* Remove dependency on jsoo. As far as I can tell, it's not used anywhere.

### DIFF
--- a/gen_js_api.opam
+++ b/gen_js_api.opam
@@ -14,8 +14,9 @@ build: [["dune" "build" "-p" name "-j" jobs]]
 depends: [
   "ocaml" {>= "4.08.0"}
   "dune" {>= "2.0"}
-  "ocaml-migrate-parsetree" {>= "1.6.0"}
+  "ocaml-migrate-parsetree" {>= "1.6.0" & < "2.0.0"}
   "ppxlib" {>= "0.9"}
+  "js_of_ocaml-compiler" { > "3.0" }
 ]
 synopsis: "Easy OCaml bindings for Javascript libraries"
 description: """
--- a/gen_js_api.opam
+++ b/gen_js_api.opam
@@ -16,6 +16,7 @@ depends: [
   "dune" {>= "2.0"}
   "ocaml-migrate-parsetree" {>= "1.6.0" & < "2.0.0"}
   "ppxlib" {>= "0.9"}
+  "js_of_ocaml-compiler" { with-test & >= "3.0" }
 ]
 conflicts: [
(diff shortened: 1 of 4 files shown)


hhugo:
Note that gen_js_api depends on js primitive provided by js_of_ocaml-compiler

mlasson:
Thanks ! Are you ok with my last commit ? If so, I'll click the merge button.

rgrinberg:
I still don't quite understand where is the jsoo compiler being used. It relies on primitive of the compiler, but that doesn't mean you need the jsoo compiler to build the library.

By the way, what do you think of splitting the runtime and ppx into separate opam packages? Still in the same repository of course.

mlasson:
> I still don't quite understand where is the jsoo compiler being used. It relies on primitive of the compiler, but that doesn't mean you need the jsoo compiler to build the library.

It's require to build the examples (which are not public) ! 
Do you think it would be better to add it with the "with-test" variable ? 
```
 "js_of_ocaml-compiler" { with-test & > "3.0" }
```

> By the way, what do you think of splitting the runtime and ppx into separate opam packages? Still in the same repository of course.

I think ... that's a good idea :).

rgrinberg:
> Do you think it would be better to add it with the "with-test" variable ?

Yes, that makes sense.

> I think ... that's a good idea :).

Will do in a separate PR.

What about switching to `(wrapped true)` in the runtime library? Will this be a difficult change for you internally?

hhugo:
@rgrinberg, gen_js_api might not be usable with older version of js_of_ocaml-compiler because it rely on javascript primitives provided by the jsoo-compiler package.

rgrinberg:
Perhaps it should be a conflict with older versions of jsoo then? That would still give us parallel package compilation.

mlasson:
> gen_js_api might not be usable with older version of js_of_ocaml-compiler because it rely on javascript primitives provided by the jsoo-compiler package

Note that I don't think this could happen. 
gen_js_api requires "ocaml >= 4.08" and should be compatible with all version of jsoo that support ocaml 4.08.

mlasson:
but sure, conflicting with js_of_ocaml < 3.0 could be more explicit.

rgrinberg:
> but sure, conflicting with js_of_ocaml < 3.0 could be more explicit.

Done.

mlasson:
May I merge ?

rgrinberg:
Yes. This is PR is ready.

PR #115, opened 2020-10-14
Support for functors and module inclusion

mlasson:
# Description 

This PR is a proposal to support functorized module during the processing of interfaces. 

I am not sure whether or not we want that. Feedback is welcomed !

# Why ? 

Sometimes binding containers as parametrized types (ie. type 'a t = ...) is complicated. 
One needs to implement their own "t_to_js = fun f x -> ..." which often consist in mapping over the whole data structure to convert from and to js. If one does that he/she has to sacrifice side-effects and all "sub-linear" performances of primitives because the conversion will need to make a copy of the data structure. 

A solution could be to use the module system in order to "pack" the to/from conversions of elements.

Of course, the result is less expressive than having a parametrized types as one will need to turn to first-class modules to implement polymorphics functions between instances of functors (eg. mapping functions) ! 

# Example 

```ocaml
module JsArray (E:
  sig
    type t
    val t_to_js: t -> Ojs.t
    val t_of_js: Ojs.t -> t
  end): sig
    type t
    val t_to_js: t -> Ojs.t
    val t_of_js: Ojs.t -> t

    val create: unit -> t [@@js.new "Array"]
    val push: t -> E.t -> unit
    val pop: t -> E.t option
end

module JsString : sig
  type t = string
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t
end

module UntypedArray : sig
  include (module type of JsArray(Ojs))
end

module StringArray : sig
  include (module type of JsArray(JsString))

  val join: t -> string -> string
end
```

# What does the PR implement ? 

Actually one can think of it as pure syntactic sugar:

## Functors

```ocaml 
module F (X : T) : sig 
  MODULE_SIGNATURE 
end
``` 
is processed as 
```ocaml
[@@@js.stop]
module F (X : T) : sig 
  MODULE_SIGNATURE  
end
[@@@js.start]
[@@@js.implem
  module F (X : T) = struct 
    include [%js: MODULE_SIGNATURE]
  end
]
```

## Module inclusion 
```ocaml
include (module type of M)
```
is processed as 
```ocaml
[@@@js.stop]
include (module type of M)
[@@@js.start]
[@@@js.implem
include M 
]
```


mlantas:
Very cool! This would be very useful for the VSCode API bindings.


```ocaml
sig
  type t
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t
end
```
Maybe the Ojs library should provide this module type? Like `Ojs.Typeable` or `Ojs.JSable`?

mlasson:
Great ! Could you give me an example ?

mlantas:
Sure, the VSCode API has a lot of classes and interfaces, so our bindings use one module per type. Some classes are generic, which are hard to write bindings for currently.

From my understanding, with this feature, we would be able to bind to the [`EventEmitter`](https://code.visualstudio.com/api/references/vscode-api#EventEmitter%3CT%3E) class like this:
```ocaml
module EventEmitter (T : sig
  type t

  val t_to_js : t -> Ojs.t

  val t_of_js : Ojs.t -> t
end) : sig
  type t = private Ojs.t

  val event : t -> Event(T).t [@@js.get]

  val fire : t -> data:T.t -> unit [@@js.call]

  val dispose : t -> unit [@@js.call]
end
```

mlantas:
It seems that with the current release we can already do the following in a .ml file with the ppx.

```ocaml
module EventEmitter (T : sig
  type t

  val t_to_js : t -> Ojs.t

  val t_of_js : Ojs.t -> t
end) =
struct
  module EventT = Event (T)

  type t = private Ojs.t [@@js]

  val event : t -> EventT.t [@@js.get]

  val fire : t -> data:T.t -> unit [@@js.call]

  val dispose : t -> unit [@@js.call]
end
```
The only part that doesn't work from the previous example is `Event(T).t`

mlasson:
Thanks for the examples ! 

> Maybe the Ojs library should provide this module type? Like `Ojs.Typeable` or `Ojs.JSable`?

It's a bit painful to add to the "Ojs" library: we need to add an alias for the toplevel "type t". 
So let's not do it :).

mlantas:
> It's a bit painful to add to the "Ojs" library: we need to add an alias for the toplevel "type t". 

Maybe I'm misunderstanding, but why would an alias be necessary? Can't the type just be kept abstract?

mlasson:
> > It's a bit painful to add to the "Ojs" library: we need to add an alias for the toplevel "type t".
> 
> Maybe I'm misunderstanding, but why would an alias be necessary? Can't the type just be kept abstract?

The problem is that you cannot put
```ocaml
module type T = sig 
  type t
  val t_to_js : t -> Ojs.t
  val t_of_js : Ojs.t -> t
end
```
inside the Ojs module without doing some kind of alias: 
```ocaml
type js  = t
module type T = sig 
  type t
  val t_to_js : t -> js
  val t_of_js : js -> t
end
```
which is ugly :/.

mlantas:
I understand now, thanks.

It's not too much work to write the module type, so I guess it's okay if it's not included in the Ojs library.

mlantas:
> which is ugly :/.

Not sure if it changes your mind, but I recently discovered ["local substitution declarations"](https://ocaml.org/releases/4.11/htmlman/signaturesubstitution.html#ss:local-substitution) which make it a little nicer:

```ocaml
module type T = sig
  type js := t
  type t
  val t_to_js : t -> js
  val t_of_js : js -> t
end
```
This only lets library users have access to the `t` type and has better editor interactions:

image

Issue #117, opened 2020-10-14
js.global with payload doesn't work with js.scope


jchavarri:
The following input:

```ml
include [%js:
module T: sig
  val log: 'a -> unit [@@js.global "jsLog"]
end [@js.scope "jsT"]
]
```

errors out with `Spurious js.* attribute`.

I would expect it to behave the same as:

```ml
include [%js:
module T: sig
  val log: 'a -> unit [@@js.global] (* Note the absence of payload here *)
end [@js.scope "jsT"]
]
```

And generate the following code (for the original input):

```ml
module T = struct
  let (log : 'a -> unit) =
   fun x1 -> ignore (Ojs.call (Ojs.get Ojs.global "jsT") "jsLog" [|Obj.magic x1|])
end
```

mlasson:
Fix by #119

Issue #116, opened 2020-10-14
Conversion functions not included in module signature


jchavarri:
Input (`.ml` file):

```ml
include [%js:
module M: sig
  type t
end
]
```

Generates (simplified):
```ml
include (
  struct
    module M = struct
      type t = Ojs.t

      let rec (t_of_js : Ojs.t -> t) = fun x2 -> x2

      and (t_to_js : t -> Ojs.t) = fun x1 -> x1
    end
  end :
    sig
      module M : sig
        type t
      end
    end )
```

One has to include the conversion functions `t_of_js` and `t_to_js` manually on the signature, which is quite tedious. Maybe gen_js_api could include them as well there?

mlasson:
Yes I known it's tedious but currently [%js: ] behaves like the standalone tool; it does not rewrite the signature at all (ie. interfaces are not preprocessed: you could compile the .mli without preprocessing it and it should be a valid interface for the generated structure).

This design decision was taken a long time ago because it was convenient to integrate with ancient build systems. 

With the advent of dune this maybe its time to change that (in particular, this may be needed to support ppxlib). 

We could do a quick fix for the [%js: extension] but I prefer to wait at least for the ppxlib integration to take a decision.

PR #118, opened 2020-10-14
Add ppx tests setup

jchavarri:
Includes a minimal setup of "snapshot"-like tests for the ppx. This kind of tests have been really helpful to me, not only to know when the output of the ppx is changing and how, but also as documentation.

I just added a couple of tests for now to cover the issues mentioned in #116 and #117, in order to ask for feedback. If this sounds like a good idea, maybe the content of `test_bindings.mli`, or a subset of it, could be added to `input.ml`.


mlasson:
Good idea ! 
I added some tests and found some bugs. 
This will be convenient to avoid regression while migrating to ppxlib.

ppx-test/issues.ml
```diff
@@ -0,0 +1,7 @@
+module Issue116 = [%js: type t]
```
jchavarri, line 1: 💯

jchavarri (commented):
Looks great @mlasson ! Good to hear it'll be useful with ppxlib migration :)

Should the `binding_*.mli` files include an example with `[@js.scope]`? Or do you prefer to add it after #117 is fixed?

Issue #112, opened 2020-08-24
calling new on a dynamic value


joprice:
Is it possible to call new on a constructor that is passed as an argument? Sometimes libraries have nested constructors that are only reachable after creating another object, and libraries loaded with "require" would not expose a global path to the constructor.

mlasson:
would you have an example of such a library ?

joprice:
I was seeing if the boilerplate of this wrapper around the electron apis could be replaced with gen_js_api: https://github.com/xvw/ocamlectron/blob/develop/lib/electron_plumbing/Struct.ml#L331. 

That wrapper specifically around BrowserWindow gets called with the constructor type that is defined here https://github.com/xvw/ocamlectron/blob/4bc80200f518a0d41a0488f6c14b27b78626ed78/lib/electron_plumbing/BrowserWindow.mli#L155, 

accessed as a property of the electron object here https://github.com/xvw/ocamlectron/blob/4bc80200f518a0d41a0488f6c14b27b78626ed78/lib/electron_plumbing/ElectronMain.mli#L9, 

which is loaded from a wrapper around "require" here https://github.com/xvw/ocamlectron/blob/1f8e4d10e5fc6ead5bfa8d9dfdc083c09c5e1c76/lib/electron_main/Electron.ml#L3

In js, it is typically imported or destructured like

```js
const {BrowserWindow} = require('electron');

const window = new BrowserWindow({...});
```

cannorin:
Now this should be possible with #145. I didn't notice this issue back then 😅

mlasson:
Let's close the issue then :) ! Thanks.

PR #119, opened 2020-10-16
Better support for global bindings & Node tests

mlasson:
# Context 

This PR allows to use js.scope to pass runtime value used to replace the "Global Object". 

It allows to reuse the js.scope logic to support "require" for nodejs app. 
In that sense it is an alternative implementation of PR#68. 

This PR also addresses the following issues: 

 - Issue #112 
 - Issue #117 
 - Issue #66 

# New Changes & Features 

  - [@js.scope] can be used to  add a scope to the "rest of a module" (which is convenient to
add a scope at the beginning of a file)

  - [@js.scope] does not change anymore the heuristics in automatic bindings (this a breaking change !)

  - [@js.scope] accept arbitrary expression of type Ojs.t (may be helpful to dynamically load library)

  - Two new modes to control the automatic bindings: [@@@js.static] and [@@@js.prototype]

  - A new kind of [@@new]: [@@create] which behave as new exception it instantiate the global object
    which is assumed to be a constructor object (I don't think it will ever make sense when not under 
    a scope !)

   - A new test directory "node_test" that gives a lot of examples and should provide a not so bad 
   non-regression testsuite. But it could be enhance of course !

   - Adds a Obj.unit_to/of_js that maps to "undefined" and lift the restriction about using unit outside 
   of functions. This can be useful to bind "unit promises" for instance but it should be handled with 
   care.
 
  - Attributes may be added directly on module declarations ! (ie. module [@attribute] M: ...) instead of module M : sig edn[@@attribute]. 
  
  - [@js.scope] may be inferred from the module name

Ps: I think I inadvertently tag the github users attribute and create :/. Sorry about that !

# Example: Errors.mli 

```ocaml
module [@js.scope] Error : sig
  type t
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t

  [@@@js.static]
  val create: string -> t
  val stack_trace_limit: int
  val set_stack_trace_limit: int -> unit

  [@@@js.prototype]
  val code: t -> string
  val message: t -> string
  val stack: t -> string
end
```

# Example: Fs.mli
```ocaml
[@@@js.scope Imports.fs_promises]

module[@js.prototype] Dirent : sig
  type t = Ojs.t
  val t_of_js: Ojs.t -> t
  val t_to_js: t -> Ojs.t

  val name: t -> string
  val is_file: t -> bool
  val is_directory: t -> bool
end

module[@js.prototype] Dir : sig
  type t = Ojs.t
  val t_of_js: Ojs.t -> t
  val t_to_js: t -> Ojs.t

  val path: t -> string
  val close: t -> unit Promise.t
  val read:t -> Dirent.t option Promise.t
end

module[@js.prototype] FileHandle : sig
  type t = Ojs.t
  val t_of_js: Ojs.t -> t
  val t_to_js: t -> Ojs.t

  type read = {
    bytes_read: int;
    buffer: Buffer.t;
  }

  val append_file: t -> Buffer.t -> unit Promise.t
  val read: t -> Buffer.t -> int -> int -> int -> read Promise.t
  val chmod: t -> int -> unit Promise.t
  val chmown: t -> uid:int -> gid:int -> unit Promise.t
  val close: t -> unit Promise.t
  val datasync: t -> unit Promise.t
  val fd: t -> int
end

[@@@js.static]
val readdir: string -> string list Promise.t
val open_: string -> flag:string -> FileHandle.t Promise.t
val rmdir: string -> unit Promise.t
val rename: string -> string -> unit Promise.t
val unlink: string -> unit Promise.t
```


node-test/bindings/path.mli
```diff
@@ -0,0 +1,19 @@
+[@@@js.scope (Ojs.require "path")]
+
+val sep: string
```
jchavarri: I checked the output from this module, and it seems only this function gets "scoped", while the rest are just translated normally.

Is this expected? From a user perspective, it's a bit confusing that `scope` might apply conditionally based on the signature of the value. I would expect `scope` to either apply to all values, or apply only if `global` is used.

lib/ojs.ml
```diff
@@ -113,3 +113,6 @@ let is_null x =
 
 let obj_type x =
   string_of_js (call (pure_js_expr "Object.prototype.toString") "call" [|x|])
+
+let internal_require = pure_js_expr "require"
+let require name = apply internal_require [| string_to_js name |]
```
jchavarri: I'm not sure if adding `require` to gen_js_api  is too specific for the library purposes (convert from/to JS). There are plenty of ways of including 3rd party modules in JavaScript. Now Node supports [`import`](https://nodejs.org/api/esm.html#esm_import_specifiers) too, browsers can load `<script>` tags, etc.

 I started writing bindings that would provide the `require` call on behalf of user, to save them some time and trouble. And quite quickly [got feedback](https://github.com/jchavarri/jsoo-react/issues/47) from folks that didn't need `require` but instead preferred loading with `<script>` tags. In the end I just removed all 3rd party loading assumptions from the library and let the users decide how they want to import JS modules.

I guess it's not a big deal, but wanted to share this piece of context for what it's worth. Maybe it's worth to add some documentation about ways to reference 3rd party JavaScript loaded libs.

As a side note, I still see the [`Critical dependency: the request of a dependency is an expression`](https://stackoverflow.com/questions/42908116/webpack-critical-dependency-the-request-of-a-dependency-is-an-expression) that I mentioned [here](https://github.com/LexiFi/gen_js_api/issues/66#issuecomment-526844888), when loading this code with Webpack when Dune builds is in `dev` profile, as it produces this output:
```js
    var internal_require=require;
    function require$0(name)
     {return internal_require(runtime.caml_js_from_string(name))}
```

jchavarri (commented):
This looks amazing @mlasson !

node-test/bindings/path.mli
mlasson: I did not change (yet !) the way [@js.scoped] affect automatic resolution which is a bit buggy (in auto mode, the implicit unscoped "get" was catching most cases. Let's fix that !

lib/ojs.ml
mlasson: agreed ! let's remove Ojs.require and use a javascript files to execute the imports in the test/example.

mlasson:
I got a bit carried away with this pull-request :). 

It introduces some breaking changes and took some decision, I would love some feedback/reviews before merging this !

VALUES.md
```diff
@@ -53,6 +53,22 @@ Supported forms
   As for global values, it is possible to indicate the access path by
   using `[@js.scope]` attributes on englobing modules (see below).
 
+  When the global object is itself an object constructor, the `[@@js.create]`
+  attribute may be used to instantiate it.
+
+  For instance,
+  ```ocaml
+    module[@js.scope "JavascriptClassName"] C : sig
+      val create: T1 -> ... -> Tn -> t [@js.create]
```
jchavarri, line 62: Would it make sense to have this be `[@js.new]`, with no payload? I guess the need to introduce a new attribute name `js.create` is to be able to tell gen_js_api where to pick the name from (value name or current global), right?

I am thinking of an alternate name `scoped_new` instead of `create` 🤔 so that the connection with `new` remains in the attribute name.

VALUES.md
```diff
@@ -226,35 +224,95 @@ Supported forms
 
   See [Verbatim section](IMPLGEN.md) for more details and examples.
 
+Scope
+-----
+
+The signature attribute `[@@@js.scope "property"]` change the reference to the current global
+object by following the property provided as payload. They can be nested.
```
jchavarri: Subjective, but imo:

> The access path is then composed by concatenation of all the names indicated by `[@js.scope]` attribute, separated by a '.'.

was more explicit and clear than

> They can be nested.

VALUES.md
```diff
@@ -226,35 +224,95 @@ Supported forms
...
+a payload to replace the global object. The intended use case is to allow
+dynamic loading of modules.
 
 Automatic binding
 -----------------
 
 Some conventions, based on the declared value names and their types,
-allow to get rid of the explicit `[@@js.xxx]` attributes on value
-declarations in most cases.  Here are the rules, applied in order:
+allow to inter implicitly the `[@@js.xxx]` attributes on value
```
jchavarri: `infer`?

VALUES.md
```diff
@@ -226,35 +224,95 @@ Supported forms
...
+There are three modes that controls the heuristic:
+  1. The *default* mode,
+  2. The *static* mode,
+  3. And the *prototype* mode.
 
-- Similarly, if the type has the form `Ojs.t -> τ` (for a local named
-  type `τ`) and the value name is `τ_of_js` (i.e. the type name
-  followed by `_of_js`), then the function is assumed to be a
-  `[@@js.cast]`.
+The signature attributes [@@@js.auto], [@@@js.prototype] and
```
jchavarri: I think `js.prototype` should be `js.static`. (`js.prototype` appears twice)

VALUES.md
```diff
@@ -226,35 +224,95 @@ Supported forms
 
   See [Verbatim section](IMPLGEN.md) for more details and examples.
 
+Scope
+-----
+
+The signature attribute `[@@@js.scope "property"]` change the reference to the current global
+object by following the property provided as payload. They can be nested.
+
+The simple use case is to binds values packed in singleton objects or classes.
```
jchavarri:
```suggestion
A simple use case is to bind to JavaScript values packed in singleton objects or classes.
```

VALUES.md
```diff
@@ -263,21 +321,62 @@ declarations in most cases.  Here are the rules, applied in order:
...
+
+- Otherwise, the declaration is assumed to be a `[@@js.global]` value.
+
+### Rules for the *prototype* mode
+
+In a *prototype* section, the generated code is not allowed to depend
+on the current global object (or an error will be emitted). Therefore
+such sections are guaranteed not to depend on the current scope.
+
+The rule in this mode are:
```
jchavarri: `rules`

VALUES.md
```diff
@@ -226,35 +224,95 @@ Supported forms
 
   See [Verbatim section](IMPLGEN.md) for more details and examples.
 
+Scope
+-----
+
+The signature attribute `[@@@js.scope "property"]` change the reference to the current global
```
jchavarri: `changes`

VALUES.md
```diff
@@ -226,35 +224,95 @@ Supported forms
...
+  module Console: sig
+    val log: string -> unit [@@js.global "console.log"]
+  end
+```
+
+When attached directly to a module, the payload of `[@@js.scope]`
+may be omitted, it will be implicitly filled with the module name
+(preserving the capitalization !).
+
+Before version 1.0.7, the presence of `[@@js.scope]` change
```
jchavarri: `changed`, or `used to change`?

VALUES.md
```diff
@@ -263,21 +321,62 @@ declarations in most cases.  Here are the rules, applied in order:
...
 
-- If the value is a function with two arguments `τ1 -> τ2 -> unit` and
+- If the value is a function whose first argument is a named type `τ
+  -> ...`, then the definition is assumed to be a `[@@js.call]` method
+  call.
+
+- Otherwise, the declaration is assumed to be a `[@@js.global]` value.
+  This applies in particular for any non-functional type.
+
+### Rules for the *static* mode
```
jchavarri: Is `static` mode strictly needed? What is it adding that the auto mode can't do?

jchavarri (commented):
This looks great @mlasson, I really like how the scope feature implementation is now much more consistent. I left some inline comments with specific feedback.

More in general, maybe I need more time to test these new features, but at first sight I am not super convinced about the user experience of the 3 modes introduced (although they seem very useful). My opinion is that for a layer as unsafe as bindings, it's better to be over-explicit and verbose rather than implicit and brief. `gen_js_api` had already a lot of "magic" going on for name mangling and other heuristics, and with the new modes, the magic is multiplying.

For example, I would prefer to write this:

```ocaml
module [@js.scope] Error : sig
  type t
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t

  val create: string -> t
  val stack_trace_limit: int
  val set_stack_trace_limit: int -> unit

  val code: t -> string [@@js.get]
  val message: t -> string [@@js.get]
  val stack: t -> string [@@js.get]
end
```

over what the example in `errors.mli` shows with modes. Even if I have to write `[@@js.get]` 3 times, it is quite obvious what each value will produce as bindings. 

With modes, the reader needs to first parse the code, then apply the gen_js_api "rules" in their head, in order to understand what bindings will be generated.

Another argument against "magic" is related to long term maintenance of gen_js_api. Adding more implicit behavior leaves less room to implement breaking changes in the future. Existing bindings libraries in "user land" could break in very subtle ways by producing different code if the heuristics were ever going to change. So users might decide to stay in older versions. On the other hand, explicit attributes are much easier to detect by gen_js_api (for example, to show deprecating warnings in specific attributes before a breaking change, etc).

But this is my opinion, maybe too conservative. And I understand gen_js_api was already "on track" to leverage these heuristics. I'm curious what you and others think?

Lupus:
This is really amazing @mlasson ! Looks neat. I can see some difficulties with long term maintenance of "mode" approach as pointed out by @jchavarri, but the bindings code reads well IMO.

Is class-level wrapping still a thing I wonder? With `prototype` mode one gets neat way to wrap JS objects in OCaml modules. The only thing that does not fit well is modelling inheritance, which I currently use for Node APIs (EventEmitter and up the hierarchy). But probably it can just be modelled as `toEventEmitter` converters annotated with `cast`.

VALUES.md
mlasson: > Would it make sense to have this be [@js.new], with no payload? I guess the need to introduce a new attribute name js.create is to be able to tell gen_js_api where to pick the name from (value name or current global), right?

In the "create" case, there is no name ! The global object is the constructor. 

Let's say you want to bind:
```js 
class Point {
  constructor(x, y) {
    this.x = x;
    this.y = y;
  }

  norm() {
    return Math.hypot(x, y);
  }

  static distance(a, b) {
    const dx = a.x - b.x;
    const dy = a.y - b.y;

    return Math.hypot(dx, dy);
  }
}
```
```ocaml
module[@js.scope "Point"] Point : sig 
  type t 
  val t_to_js: t -> Ojs.t
  val t_of_js: Ojs.t -> t 

  val new_point: float -> float -> t [@@js.create] (* no way to bind the constructor using [@js.new] under the scope *)
  val distance: t -> t -> float [@@js.global] 
  
  val norm: t -> float [@@js.call]
end 
```

If we were to use `[@@js.new]` without payload, it would be the same as [@@js.new "Point"] (which under the scope "Point" would be "Point.Point").

The problem with "scoped_new" is that it could imply that "new" is "unscoped". We could use the empty string "" instead of no payload (`[@@js.new ""]` ?) Or `[@@js.new_self]` ?

mlasson:
> More in general, maybe I need more time to test these new features, but at first sight I am not super convinced about the user experience of the 3 modes introduced (although they seem very useful). My opinion is that for a layer as unsafe as bindings, it's better to be over-explicit and verbose rather than implicit and brief. gen_js_api had already a lot of "magic" going on for name mangling and other heuristics, and with the new modes, the magic is multiplying.

Ok, I agree. Let's get rid of these two modes. 

> Is class-level wrapping still a thing I wonder? With prototype mode one gets neat way to wrap JS objects in OCaml modules. The only thing that does not fit well is modelling inheritance, which I currently use for Node APIs (EventEmitter and up the hierarchy). But probably it can just be modelled as toEventEmitter converters annotated with cast.

Well, you got it right. Caml objects have a better story wrt to inheritance but it adds a thin "extra" layer of caml into your bindings which may feel unnecessary. In practice, we often use "Ojs.t" to bind event target and do explicit to/of_js (or cast) in situation we know what kind of event we are expecting. I would love to see a more "type-safe" interfaces to events APIs.

PR #120, opened 2020-10-16
Split runtime library to own package

rgrinberg:
First, I moved the opam file generation to dune. This makes it easier to
maintain multiple opam files.

In this PR, the runtime package is now in `ojs`, to match the library name.
Other splits are possible, but I think they're more intrusive:

* gen_js_api_runtime & gen_js_api
* gen_js_api & gen_js_api-ppx

Should we create an alias for the old library name gen_js_api? This will
maintain compatibility. Or `gen_js_api` could just be the ppx? This seems like a
nicer convention.

Should `ojs` be included in the `ppx_runtime_libraries` for `gen_js_api.ppx`?
This will mean that users will no longer need to `ojs` as a dependency manually.


dune-project
```diff
@@ -1,2 +1,39 @@
...
+
+gen_js_api is to be used with the js_of_ocaml compiler.
+ ")
+ (conflicts (js_of_ocaml-compiler (< 3.0.0)))
+ (depends
+  (ocaml (>= 4.08))
+  (ppxlib (>= 0.9))
+  (ocaml-migrate-parsetree (and (>= 1.6.0) (< 2.0.0)))
+  (js_of_ocaml-compiler :with-test)
+  ojs))
```
mlasson, line 39: @rgrinberg Do you think we should add the runtime library "ojs" as an explicit dependency to the ppx ?

mlasson:
> Should ojs be included in the ppx_runtime_libraries for gen_js_api.ppx?

Done.

@rgrinberg If you are ok with my modifications, I'll merge this.

dune-project
rgrinberg: Yes, I think so. Did I accidentally omit it?

mlasson: No ! But I was not sure it was required.

rgrinberg:
Looks good. Thanks for the fixes!

mlasson:
Thank you !

PR #125, opened 2020-12-13
js.custom attribute for type declaration to support custom mapping

### LINKED ISSUE #124: [question] How to define mutually recursive types including both JS-able types and non-JS-able types (with custom `of_js` and `to_js`)?
Assume we want to define two types `t` and `u` with the following definitions:

```ocaml
[@@@js.stop]
type t = (* alias to some not JS-able type *)
val t_to_js: t -> Ojs.t
val t_of_js: Ojs.t -> t
[@@@js.start]
[@@@js.implem
  type t = (* alias to some not JS-able type *)
  let t_to_js x = (* some code *)
  let t_of_js x = (* some code *)
]
```

```ocaml
type u =
  | Unknown of Ojs.t [@js.default]
  | T of t [@js "t"]
  [@@js.union on_field "type"]
```

If t is an alias to some type that involves `u`, `t` and `u` becomes mutually recursive. But if the said type is not supported by gen_js_api (i.e. using poly variants without `js.*` attribute), things become difficult to make it compile.

Now neither of the followings work:

```ocaml
type 'a dummy

type u =
  | Unknown of Ojs.t [@js.default]
  | T of t [@js "t"]
  [@@js.union on_field "type"]

and t = [`U of u] dummy
(* Error: Sum types without js.* attribute not supported in this context *)
```

```ocaml
type 'a dummy

type u =
| Unknown of Ojs.t [@js.default]
| T of t [@js "t"]
[@@js.union on_field "type"]

[@@@js.stop]
and t = [`U of u] dummy (* Error: Syntax error *)
[@@@js.start]
[@@@js.implem
and t = [`U of u] dummy (* Error: Syntax error *)
]
```
The best possible workaround I was able to invent is:

```ocaml
type 'a dummy

[@@@js.stop]
type 'u tmp = [ `U of 'u ] dummy
val tmp_to_js: ('u -> Ojs.t) -> 'u tmp -> Ojs.t
val tmp_of_js: (Ojs.t -> 'u) -> Ojs.t -> 'u tmp
[@@@js.start]
[@@@js.implem
  type 'u tmp = [ `U of 'u ] dummy
  let tmp_to_js _ x = Obj.magic x
  let tmp_of_js _ x = Obj.magic x
]

type u =
| Unknown of Ojs.t [@js.default]
| T of t [@js "t"]
[@@js.union on_field "type"]

and t = u tmp
```

But this requires a bit too much of coding, loses the information of `u` when writing custom `to_js`/`of_js` functions, and introduces an unwanted type `tmp` to the scope.

Is there some way to solve this problem nicely? I think it would be great if an attribute (say `js.ignore`) exists to let the type definition accepted without being processed by gen_js_api so that the following compiles:

```ocaml
type 'a dummy

type u =
  | Unknown of Ojs.t [@js.default]
  | T of t [@js "t"]
  [@@js.union on_field "type"]

and t = [`U of u] dummy [@@js.ignore]

[@@@js.stop]
val t_to_js: t -> Ojs.t
val t_of_js: Ojs.t -> t
[@@@js.start]
[@@@js.implem
  let t_to_js x = (* some code *)
  let t_of_js x = (* some code *)
]
```

cannorin:
Closes #124 .

This PR introduces an ability to provide custom mapping function to a type declaration in `.mli` files.

This makes it much easier to define mutually recursive types including a type not supported by gen_js_api.

```ocaml
  type 'a dummy

  type u =
    | Unknown of Ojs.t [@js.default]
    | T of t [@js "t"]
    [@@js.union on_field "type"]

  and t = [`U of u] dummy [@@js.custom
    let t_to_js x = Obj.magic x
    let t_of_js x = Obj.magic x
  ]
```

The above produces the following output:

```ocaml
      type 'a dummy = Ojs.t
      let rec (dummy_of_js : (Ojs.t -> 'a) -> Ojs.t -> 'a dummy) =
        fun _ -> fun x7 -> x7
      and (dummy_to_js : ('a -> Ojs.t) -> 'a dummy -> Ojs.t) =
        fun _ -> fun x6 -> x6
      type u =
        | Unknown of Ojs.t [@js.default ]
        | T of t [@js "t"][@@js.union on_field "type"]
      and t = [ `U of u ] dummy[@@js.custom
                                 let t_to_js x = Obj.magic x
                                 let t_of_js x = Obj.magic x]
      let rec (u_of_js : Ojs.t -> u) =
        fun x11 ->
          let x12 = x11 in
          match Ojs.type_of (Ojs.get x12 "type") with
          | "number" ->
              (match Ojs.int_of_js (Ojs.get x12 "type") with
               | _ -> Unknown x12)
          | "string" ->
              (match Ojs.string_of_js (Ojs.get x12 "type") with
               | "t" -> T (t_of_js x12)
               | _ -> Unknown x12)
          | _ -> Unknown x12
      and (u_to_js : u -> Ojs.t) =
        fun x8 -> match x8 with | Unknown x9 -> x9 | T x10 -> t_to_js x10
      and t_to_js x = Obj.magic x
      and t_of_js x = Obj.magic x
```

This PR also includes a minimum test and some documentation updates.

TODO:
- [x] Make CI green
- [ ] More docs?
- [ ] More tests?


ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1284,8 +1287,43 @@ and gen_funs ~global_attrs p =
...
+                  | Pstr_include incdec ->
+                    let result =
+                      match incdec.pincl_mod.pmod_desc with
+                      | Pmod_structure s' -> get_value_bindings s'
+                      | _ -> []
+                    in result
+                  | _ -> []
+                ) s
+              in
+              let vbs = get_value_bindings structure in
```
alainfrisch: It feels a bit weird to inject any value declaration (toplevel or under an included literal struct -- is that case useful?) into an implicitly recursive binding, and ignore (without complaining) all others structure items.

Another direction would be to use a more structured payload, e.g a record with two fields:

````ocaml
[@@js.custom { of_js = (fun ... -> ...); to_js = (fun ... -> ...) }]
````
(in your example:

````ocaml
[@@js.custom { of_js = Obj.magic; to_js = Obj.magic }]
````

I'm not opposed to keeping the current approach (with two value bindings), but then I'd prefer failing explicitly on any other structure item.  Or do see cases where allowing extra value bindings  can be useful?

File: ppx-lib/gen_js_api_ppx.ml [hunk 2]
```diff
@@ -1284,8 +1287,43 @@ and gen_funs ~global_attrs p =
...
+              assert_function (name ^ "_of_js") vbs;
+              assert_function (name ^ "_to_js") vbs;
+              (fun _ -> true),
+              lazy (raise (Error (loc, Union_without_discriminator))),
```
alainfrisch: I don't have the current logic fresh in my mind, but this looks suspicious : is that lazy ever going to be forced?

alainfrisch (commented):
Thanks for the contribution!

I like the new feature, but I'm not fully convinced by the design of the payload for the new attribute; see inline comment. To be discussed!

ppx-lib/gen_js_api_ppx.ml
cannorin: Yes. It's basically to disable the automatic generation of `_to_js` and `_of_js`. See: https://github.com/LexiFi/gen_js_api/blob/17afcd648b1616badcbaebef561c0a445eaa6641/ppx-lib/gen_js_api_ppx.ml#L1372-L1373

We already have `value_binding`s defined by user, so we don't want to use `push_fun` and just use them as they are.
https://github.com/LexiFi/gen_js_api/blob/17afcd648b1616badcbaebef561c0a445eaa6641/ppx-lib/gen_js_api_ppx.ml#L1406-L1410

But the name of the exception we are using might be confusing. It might be better to add a new one.

ppx-lib/gen_js_api_ppx.ml
cannorin: I used a structure here because we can get the `value_binding`s directly from the `PStr`.  If we used a record here, we have to manually convert `Pexp_fun` to `value_binding`, which can be a little pain. I think it's easier to just reject any structure items other than `_to_js` and `_of_js`.

cannorin: Implemented with https://github.com/LexiFi/gen_js_api/pull/125/commits/b95f420e92c6912c2a15fa396b7038f816dead83

ppx-lib/gen_js_api_ppx.ml
alainfrisch: Ah, yes.  It's better to introduce another exception constructor for that purpose.

ppx-lib/gen_js_api_ppx.ml
alainfrisch: Sorry for the late reply.  I still find it misleading that (1) `let t_of_js = ...  let t_to_js = ...` and (2) `let rec t_of_js = ... and t_to_js = ...` mean the same thing.

If you are concerned with the implementation complexity, I suggest you try with the syntax `{of_js=...; to_js=...}`; it should really not be more complex to implement.

cannorin:
Rebased to master to reflect changes from #127.

ppx-lib/gen_js_api_ppx.ml
cannorin: Okay, let's try `{of_js=...; to_js=...}` then.

cannorin: Implemented. Please review it 🙏

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1323,39 +1323,60 @@ and gen_funs ~global_attrs p =
...
+                    in
+                    let expected_names = [ "of_js"; "to_js" ] in
+                    let field_to_value_binding
+                      (nameloc: Longident.t with_loc)
+                      (body: expression)
+                      (remaining_names: label list) : value_binding * label list =
+                      match nameloc.txt with
+                      | Lident label ->
+                        let nameloc = { txt = Printf.sprintf "%s_%s" name label; loc = nameloc.loc } in
+                        let pat =
```
alainfrisch: You could use `Ast_helper` to make this code a bit simpler (`Pat.var ~loc:nameloc.loc nameloc`, ...).

mlasson:
I did the simplification regarding using Ast_helper, also I added further small simplification regarding the parsing of the payload. 
@cannorin : do not hesitate to object !

cannorin:
@mlasson I'm perfectly ok with these changes :+1:

PR #126, opened 2020-12-22
Put Ojs_exn inside Ojs module (breaking change !)

### LINKED ISSUE #121: Change Ojs_exn to Ojs.Exn
It's recommended that every library has a single toplevel module, and `(wrapped false)` is only there to help people transition to dune. Would it be ok to break compatibility and to switch to `Ojs.Exn`? If not, then perhaps we could add something to dune to help the transition from `(wrapped false)` to `(wrapped true)`. We already have one such mechanism `(wrapped (transition "message"))` but it doesn't quite work in some cases (like this one)

mlasson:
This module does not seem to be used by public libraries on opam.
I don't think this will break a lot of foreign code.


alainfrisch:
We need to be careful, since forcing to link Ojs_exn makes it more difficult use gen_js_api together with js_of_ocaml stdlib or [brr](https://github.com/dbuenzli/brr). This is discussed in https://github.com/dbuenzli/brr/issues/2.

The best solution would probably be to have js_of_ocaml come with a "core" interop stdlib that registers the exception to the runtime system, and possibly expose primitives of the runtime system, allowing higher level FFI and binding system to live well together.

Short of having something like that, it seems too dangerous to force linking Ojs_exn.  One possibility to "fix" #126 could be to have Ojs_exn be its own library, but it's a bit silly.  Anyway, I don't think this PR should be merged, so I'm in favor of closing it.  @mlasson, do you agree ?

mlasson:
Yes, I was arriving at the same conclusion.

PR #128, opened 2021-01-14
Bucklescript -> ReScript (also add genType ppx as a resource)

ryyppy:
Hey,

Just wanted to update some resources. I also think that genType is a pretty cool tool, so maybe worth mentioning here.


README.md
```diff
@@ -87,8 +87,8 @@ Related projects
     [DefinitelyTyped](https://github.com/borisyankov/DefinitelyTyped)
     interfaces and produce OCaml interfaces.
 
-  - [Bucklescript](https://github.com/bloomberg/bucklescript):
-    Another compiler from OCaml to Javascript.
+  - [ReScript](https://github.com/rescript-lang/rescript-compiler):
+    Another compiler from OCaml to Javascript, featuring the [genType](https://github.com/reason-association/genType) ppx for generating TS / Flow types and runtime converters.
```
smorimoto: It's a minor point, but it might be a good idea to fix it at the same time.
`Javascript` -> `JavaScript`

smorimoto: Oh no, it's wrong in more places than I thought. I will open another PR.

ryyppy: Feel free to edit as you like!

smorimoto: I'm not a developer working at LexiFi, and I'm not a collaborator, so I don't have edit privileges for this PR. So I will deal with this with another PR!

mlasson:
Thanks !

PR #129, opened 2021-01-14
Add GitHub Actions workflow

smorimoto:
https://github.com/smorimoto/gen_js_api/actions/runs/484917182

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,51 @@
+name: Main workflow
+
+on:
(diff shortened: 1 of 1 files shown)


smorimoto:
The test on Windows is currently failing, but it has nothing to do with this PR, so we can ignore that.

alainfrisch:
Thanks!

smorimoto:
@alainfrisch You're welcome. By the way, when you have time, could you take a look at the PRs opened by cannorin?

PR #131, opened 2021-01-29
Register js as a deriver

### LINKED ISSUE #116: Conversion functions not included in module signature

mlasson:
# Context 

This PR allows to user [@@deriving js] to generate the conversion functions. 
This is helpful to avoid writing their signature in interfaces (see issue #116) but it integrates better in the ecosystem. 

# Example 

## Before 
In signature:
```ocaml 
 type ('a, 'b) t = { x : 'a; y : 'b }
 val t_of_js : (Ojs.t -> 'a) -> (Ojs.t -> 'b) -> Ojs.t -> ('a, 'b) t
 val t_to_js : ('a -> Ojs.t) -> ('b -> Ojs.t) -> ('a, 'b) t -> Ojs.t
```
and in structure : 
```ocaml 
 type ('a, 'b) t = { x : 'a; y : 'b } [@js] 
```

## After
In both signture and structure: 
```ocaml 
 type ('a, 'b) t = { x : 'a; y : 'b } [@@deriving js]
```
This attributes works in preprossed ml/mli files. 
It also works through the "standalone" tool although the resulting implementation file will still need to be preprocessed by the ppx.

### DIFF
--- a/ppx-driver/gen_js_api_ppx_driver.ml
+++ b/ppx-driver/gen_js_api_ppx_driver.ml
@@ -21,12 +21,12 @@ module To_ppxlib = struct
     | _ -> assert false
 
   let copy_attribute (a : From_ppx.Ast.Parsetree.attribute)
-  : Ppxlib.Ast.attribute =
-  let pat : Migrate_parsetree.Ast_411.Parsetree.pattern =
-    Migrate_parsetree.Ast_411.Ast_helper.Pat.any ~attrs:[a] ()
-  in
-  let pat = copy_pattern pat in
-  List.hd pat.ppat_attributes
+    : Ppxlib.Ast.attribute =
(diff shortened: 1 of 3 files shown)


mlasson:
I'm closing this as I don't think we want to do that.

PR #132, opened 2021-01-30
Add support for indexers and "callable" objects

cannorin:
This PR adds three attributes for value binding: `[@@js.index_get]`, `[@@js.index_set]`, and `[@@js.apply]`.

## Indexers

 `[@@js.index_get]` and `[@@js.index_set]` can be used to bind to indexer in a JS object. For example, the binding for getter and setter of JS `Buffer` can now be written as

```ocaml
val get: t -> int -> int option [@@js.index_get]
val set: t -> int -> int -> unit [@@js.index_set]
```

instead of

```ocaml
val get: t -> int -> int option[@@js.custom
  let get buf index = [%js.to:int option](Ojs.array_get buf index)
    ]
val set: t -> int -> int -> unit[@@js.custom
  let set buf index value =
    Ojs.array_set buf index ([%js.of:int] value)
  ]
```

## Function Objects

Sometimes a JS object can also be a function (so-called "callable object").

```js
var foo = function() { .. }
foo.bar = 42
```

We can now bind this kind of object with the following:

```ocaml
module Foo : sig
  type t = private Ojs.t

  val foo: t -> int [@@js.get "foo"]
  val apply: t -> unit -> unit [@@js.apply]
end
```

## Compatibility with TypeScript

This complements gen_js_api's ability to bind to TypeScript interfaces. With this PR merged, we now support [Function Types](https://www.typescriptlang.org/docs/handbook/interfaces.html#function-types) and [Indexable Types](https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types) (which were very hard to bind).

```typescript
interface MapLike<T> {
    [index: string]: T
}
interface SearchFunc {
    (source: string, subString: string): boolean;
}
```

```ocaml
module MapLike : sig
  type 'a t = private Ojs.t
  val get: 'a t -> string -> 'a option [@@js.index_get]
  val set: 'a t -> string -> 'a -> unit [@@js.index_set]
end
module SearchFunc : sig
  type t = private Ojs.t
  val apply: t -> string -> string -> bool [@@js.apply]
end
```


lib/ojs.mli
```diff
@@ -61,6 +61,16 @@ external get: t -> string -> t = "caml_js_get"
 
 external set: t -> string -> t -> unit = "caml_js_set"
 
+val get_with: ('a -> t) -> t -> 'a -> t
```
alainfrisch: What about exposing directly internal_{get,set}.  It doesn't really lead to simpler code on the call site to pass the "index_to_js" argument than to apply it directly.

VALUES.md
```diff
@@ -296,6 +339,10 @@ The rules are applied in order:
   `[@@js.set]` global setter (whose name is obtained by dropping the
   `set_` prefix).
 
+- If the value is a function returning `unit` with three arguments
+  whose first argument is a named type `τ -> τ2 -> τ3 -> unit`, then
+  the declaration is assumed to be a `[@@js.index_set]` index setter.
```
alainfrisch: "... and the name is `set` ..." ?

lib/ojs.mli
cannorin: Yes, let's do that.

VALUES.md
cannorin: ahh I forgot to mention that, thanks 😅

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -1594,13 +1594,7 @@ and gen_def ~global_object loc decl ty =
     Arrow {ty_args = {lab=Arg; att=_; typ} :: ty_args; ty_vararg; unit_arg; ty_res} ->
       let formal_args, concrete_args = prepare_args [] ty_args ty_vararg in
       let res this = ojs_call_arr (ml2js [] typ this) (str s) concrete_args in
-      mkfun
-        (fun this ->
-           match ty_args, ty_vararg, unit_arg with
-           | [], None, false -> js2ml_unit [] ty_res (res this)
```
mlasson, line 1573: I know you were not responsible for that; but this pattern matching was not making a lot of sense: 
`ty_vars, ty_avarg = [], None` implies `formal_args = []` implies `func formal_args false x = x`

mlasson:
Nice PR thanks !

PR #135, opened 2021-02-04
UTF-8 support for (Ojs.get/set/delete) adaptions

### LINKED ISSUE #104: Planning to change the semantic of MlBytes.toString in js_of_ocaml

mlasson:
See Issue #104. 

The change in semantics happened when ocsigen/js_of_ocaml#997 was merged but they never added the primitive proposed in #104. It means that since 3.7.0 "Ojs.get/set/delete" is broken when applied to non ascii keys (which is not the end of the world; I am not aware of any API with non ascii characters...). 

I propose a new solution to be able to close the issue: I deprecated these functions and introduced : 

```ocaml
external get_prop_ascii: t -> string -> t = "caml_js_get"
  (** Get the property from an object (only works if the property key is a plain ascii string). *)

external set_prop_ascii: t -> string -> t -> unit = "caml_js_set"
  (** Set an object property (only works if the property key is a plain ascii string). *)

external delete_prop_ascii: t -> string -> unit = "caml_js_delete"
  (** Delete an object property (only works if the property key is a plain ascii string). *)

external get_prop: t -> t -> t = "caml_js_get"
  (** Get the property from an object. *)

external set_prop: t -> t -> t -> unit = "caml_js_set"
  (** Set an object property. *)

external delete_prop: t -> t -> unit = "caml_js_delete"
  (** Delete an object property. *)
```
The ppx now use the ascii version only on ascii "properties" and fallback to "proper UTF-16 encoding" when needed (through string_to_js !).


lib/ojs.mli
```diff
@@ -57,20 +57,38 @@ external fun_to_js_args: (t -> 'a) -> t = "caml_ojs_wrap_fun_arguments"
...
+
+external delete_prop: t -> t -> unit = "caml_js_delete"
+  (** Delete an object property. *)
 
 external obj: (string * t) array -> t = "caml_js_object"
 
 val empty_obj: unit -> t
 
 val has_property: t -> string -> bool
-external iter_properties: t -> (string -> unit) -> unit = "caml_ojs_iterate_properties"
```
mlasson: I removed this function because using it is usually a bad idea and it is buggy on UTF-8 strings (independently of the change in semantics).

examples/test/main.ml
```diff
@@ -198,24 +196,11 @@ let () =
 
 (* Custom mapping between association lists and JS objects *)
 
-module Dict : sig
+module Dict = [%js:
```
alainfrisch: IIUC, this completely changes the behavior of Dict.t.  The assoc list is no longer represented as a JS object.

lib/ojs.mli
```diff
@@ -83,26 +101,17 @@ external apply: t -> t array -> t = "caml_js_fun_call"
 external new_obj: t -> t array -> t = "caml_js_new"
 (** Call a constructor *)
 
-val call_arr: t -> string -> t -> t
```
alainfrisch: Is the removal of those function related to the PR?

lib/ojs.mli
alainfrisch: The function could be useful for debugging purposes, and to bind to JS objects that represent generic association lists.  What is the current implementation buggy?

alainfrisch: In particular, it goes through caml_js_to_string, which should be ok, no?

mlasson: > The function could be useful for debugging purposes, and to bind to JS objects that represent generic association lists. 

I'm ok with debugging purposes and super sceptic about using JS objects to represent generic association lists.

But Yes, there is not bug; I misread the implementation so I will revert this change and its consequences.

lib/ojs.mli
mlasson: I could keep them there (even though the ppx is not using them anymore) (but to keep I would need to replace the use of "get" by "get_prop  + string_to_js")

PR #137, opened 2021-02-20
Add explicit polymorphic type annotations to the mapping functions of parametrized types

cannorin:
Related: #124 #125

The following declaration fails on the current version of gen_js_api:

```ocaml
  type 'a dummy

  type ('a, 'b) base = [ `BaseA of 'a | `BaseB of 'b ] dummy [@@js.custom {
    to_js = (fun _ _ -> Obj.magic);
    of_js = (fun _ _ -> Obj.magic)
  }]
  and base1 = (int, string) base
  and base2 = (string, int) base
```

because it introduces a polymorphic recursion:

```
38 |   fun x20 -> base_of_js Ojs.string_of_js Ojs.int_of_js x20 
                             ^^^^^^^^^^^^^^^^                   
Error: This expression has type Ojs.t -> string               
       but an expression was expected of type Ojs.t -> int    
       Type string is not compatible with type int                   
```

This PR fixes this by adding explicit polymorphic type annotations to the generated mapping functions if the type is parameterized.


cannorin:
hmm, the CI is failing for the older versions of OCaml... not sure how to fix this.

mlasson:
@cannorin may I merge this ?

cannorin:
@mlasson yes, please 🙂

PR #141, opened 2021-02-25
Make gen_js_api testable on older distributions

kit-ty-kate:
distributions such as Ubuntu 16.04 are old enough to not have the `node` binary but only `nodejs`


mlasson:
Thanks for the proposal but we will not do that to support old node versions that won't be able successfully run the tests anyway !

PR #143, opened 2021-03-09
Disable eta reduction for of_js and to_js of type aliases

### LINKED ISSUE #142: Recursive type definitions generate invalid `let-rec` definitions
Consider the following code:

```ocaml
module Issue142 = [%js:
  type t = [`Foo [@js 42]] [@js.enum]
  and u = t
]
```

This would generate the following implementation:

```ocaml
module Issue142 : sig type t = [ `Foo ]
                      and u = t end =
  ((struct
      [@@@js.dummy "!! This code has been generated by gen_js_api !!"]
      [@@@ocaml.warning "-7-32-39"]
      type t = [ `Foo ]
      and u = t
      let rec t_of_js : Ojs.t -> t =
        fun (x40 : Ojs.t) ->
          let x41 = x40 in
          match Ojs.int_of_js x41 with | 42 -> `Foo | _ -> assert false
      and t_to_js : t -> Ojs.t =
        fun (x39 : [ `Foo ]) -> match x39 with | `Foo -> Ojs.int_to_js 42
      and u_of_js : Ojs.t -> u = t_of_js
      and u_to_js : u -> Ojs.t = t_to_js
    end)[@merlin.hide ]) 
```

but this fails with an error ```Error: This kind of expression is not allowed as right-hand side of `let rec'``` because the right hand side of `u_of_js` is `t_of_js`, which is bound in the same group of `let rec` definitions.

cannorin:
Close #142 .

This PR disables eta reduction `(fun x -> f x) => f` on generating mapping functions `t_of_js / t_to_js` when `t` is a type alias. This prevents gen_js_api from generating invalid `let rec` definitions when a type and a type alias to the type is defined recursively.


mlasson:
Thanks !

PR #144, opened 2021-03-15
Disable "Spurious js.* attribute" error for [@js.dummy]

cannorin:
gen_js_api is currently detecting false "spurious js.* attribute" errors for the `[@js.dummy]` attributes used to define a function returning a function:

```ocaml
t1 -> (t2 -> t3 [@js.dummy])
```

When we put the following to a `.mli`:
```ocaml
module Issue144: sig
  type t
  val f: t -> (args:int -> int [@js.dummy]) [@@js.call "f"]
end
```
It fails with the error:
```
File "issues_mli.mli", line 3, characters 33-41:                                                                              
3 |   val f: t -> (args:int -> int [@js.dummy]) [@@js.call "f"]                                                                
Error: Spurious js.* attribute
```
This behavior is clearly undesirable since without the attribute it would be parsed as an arity 2 function instead of 1.

Interestingly, it doesn't reproduce when we put the following to a `.ml` instead:

```ocaml
module Issue144 = [%js:
  type t
  val f: t -> (args:int -> int [@js.dummy]) [@@js.call "f"]
]
```

I suspect the way gen_js_api detects spurious attributes might be broken, but I couldn't figure out why. This PR is just a temporary fix for this particular problem.


mlasson:
Thanks !

cannorin:
Sorry, forgot to update the CHANGES.md...

tmattio:
I still see this error in `1.0.8`:

```
556 |     val bind''''': t -> this:(this:'T -> args:('AX list [@js.variadic]) -> 'R) -> thisArg:'T -> args:('AX list [@js.variadic]) -> (args':('AX list [@js.variadic]) -> 'R [@js.dummy]) [@@js.call "bind"]
                                                                                                                                                                                 ^^^^^^^^
Error: Spurious js.dummy attribute
```

Is the fix above supposed to apply here as well?

mlasson:
Nice corner case ! A variadic function returning a variadic function. 
Well, in that case the dummy is not required (and thus spurious !) because it is always considered to be a higher-order function (because if it was not higher-order we would not know how to "split" the list of arguments between the two lists). 

In a future version, we may do something to accept the "useless" dummy there if it is a problem.

tmattio:
Great, thanks for the quick answer 🙂

cannorin:
Can't we just ignore every `js.dummy` attribute for the spuriousness check? The current situation is not very helpful for automated code generation (since you have to be careful for every corner case like this).

mlasson:
@cannorin; I agree.  Could you make a small issue so I don't forget about it ?
I'll try to deal with it promptly.

PR #145, opened 2021-03-16
Add support for "newable" functions to [@@js.apply ]

cannorin:
This PR extends `[@@js.apply]` so that it can be used to bind to "newable" (must be called with a prefix `new`) functions.

This will be useful when binding to "constructor interfaces" in TypeScript:

```typescript
interface FooConstructor {
  new () : Foo
}

declare var Foo: FooConstructor;
```

```ocaml
module FooConstructor: sig
  type t
  val new_: t -> Foo.t [@@js.apply as_constructor]
end

val foo: FooConstructor.t [@@js.global "Foo"]
```

### DIFF
--- a/VALUES.md
+++ b/VALUES.md
@@ -44,6 +44,15 @@ Supported forms
 
   The name of the function need not necessarily be `apply` as long as the `[@@js.apply]` attribute is present.
 
+  When the function you want to bind is a "newable" one (a function that must be called with a prefix `new`, e.g. constructors), add `as_constructor` to the payload. This is especially useful to bind to constructor interfaces in TypeScript.
+
+  ```ocaml
+  module FooConstructor: sig
+    type t
+    val new_: t -> Foo.t [@@js.apply as_constructor]
+  end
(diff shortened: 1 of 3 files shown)


cannorin:
`as_newable` might be a better keyword than `as_constructor`. What do you think?

mlasson:
What about `[@@js.apply_constructor]` or `[@@js.instance]` ?

cannorin:
`[@@js.instance]` would be difficult for users to imagine that it is meant to bind to newable functions. `[@@js.apply_constructor]` is good, but I'm wondering `[@@js.apply_newable]` might be better to explain the actual functionality.

cannorin:
OK, let's go with `[@@js.apply_constructor]`.

mlasson:
Sorry for being late, I agree "apply_newable" is great. 
Would you mind changing it again and update the doc as well ?
Then add an entry in Changes.md and I'll merge !

cannorin:
Done 🙂

cannorin:
Sorry, forgot to update the CHANGES.md...

mlasson:
No worries ! I'll add it manually.

PR #146, opened 2021-03-16
Fix an edge-case bug of prepare_args

cannorin:
This PR fixes an edge-case bug of `prepare_args` where gen_js_api fails with `assert false` when trying to parse an arrow type with an optional parameter of some enum type like below:

```ocaml
val f: ?arg:([`Foo [@js 42]] [@js.enum]) -> unit -> int [@@js.global "f"]
```


mlasson:
Thanks !

PR #149, opened 2021-03-24
Stop using OMP directly

### LINKED ISSUE #138: Resolve the ocaml-migrate-parsetree < "2.0.0" constraint
It prevents gen_js_api from being installed at the same time as ocamlformat in a 4.12 switch, since ocamlformat since 0.17 has a ocaml-migrate-parsetree >= "2.1.0" constraint.

mlasson:
# Description

Let's stop using OMP directly and use the more idiomatic interfaces of Ppxlib. 
This allows to drop the conflict with newer version of OMP.

### DIFF
--- a/ppx-driver/gen_js_api_ppx_driver.ml
+++ b/ppx-driver/gen_js_api_ppx_driver.ml
@@ -11,8 +11,6 @@ let () =
   then (
     Ppxlib.Driver.enable_location_check ()
   );
-  Gen_js_api_ppx.mark_as_handled_manually := (fun attribute ->
-      Ppxlib.Attribute.mark_as_handled_manually attribute);
   let mapper_for_sig =
     Gen_js_api_ppx.mark_attributes_as_used
   in
--- a/ppx-driver/gen_js_api_ppx_driver.ml
+++ b/ppx-driver/gen_js_api_ppx_driver.ml
@@ -47,7 +47,6 @@ let () =
   Gen_js_api_ppx.mark_as_handled_manually := (fun attribute ->
       Ppxlib.Attribute.mark_as_handled_manually attribute);
   let mapper_for_sig =
-
     Gen_js_api_ppx.mark_attributes_as_used
   in
   let mapper_for_str =
@@ -55,10 +54,7 @@ let () =
   in
   let module_expr_ext =
     let rewriter ~loc ~path:_ si =
-      si
-
-      |> Gen_js_api_ppx.module_expr_rewriter ~loc ~attrs:[]
-
+      Gen_js_api_ppx.module_expr_rewriter ~loc ~attrs:[] si
     in
     Ppxlib.Extension.declare "js"
       Ppxlib.Extension.Context.Module_expr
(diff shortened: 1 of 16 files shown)


nojb:
There is a lot of whitespace in the diff -- is this somehow related to the switch to ppxlib ?

mlasson:
No that's the reindent commit. Please ignore whitespaces while reviewing !

mlasson:
Ok, Ok, I'll remove the commit and I'll push it after merging.

ppx-driver/gen_js_api_ppx_driver.ml
```diff
@@ -44,22 +12,16 @@ let () =
     Ppxlib.Driver.enable_location_check ()
   );
   Gen_js_api_ppx.mark_as_handled_manually := (fun attribute ->
-      let attribute = To_ppxlib.copy_attribute attribute in
       Ppxlib.Attribute.mark_as_handled_manually attribute);
   let mapper_for_sig =
-    To_ppxlib.copy_mapper
-      (Gen_js_api_ppx.mark_attributes_as_used Gen_js_api_ppx.mapper)
```
mlasson, line 51: I think this is wrong.

mlasson: Actually it seems that the recursive use of the "Gen_js_api_ppx.mapper" was not used.

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -120,7 +126,7 @@ let get_string_attribute_default key default attrs =
   | None -> default
   | Some payload -> payload.attr_loc, id_of_expr (expr_of_payload payload)
 
-let print_error ppf = function
+let _print_error ppf = function
```
nojb: How come this is not used anymore?

mlasson: Oh I forgot to deal with that :); silly me.

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -181,7 +181,12 @@ let _print_error ppf = function
 let () =
   Location.Error.register_error_of_exn
     (function
-      | Error (loc, _err) -> Some (Location.Error.make  ~loc ~sub:[] "ERROR")
+      | Error (loc, err) ->
+        let createf ~loc fmt =
+          Format.kasprintf
+            (fun str -> Location.Error.make ~loc ~sub:[] str) fmt
+        in
+        Some (createf ~loc "%a" print_error err)
```
mlasson, line 188: See https://github.com/ocaml-ppx/ppxlib/pull/239

hhugo:
You should be able to remove the ref `mark_as_handled_manually`. It was introduced when adding compatibility with ppxlib while leaving the ppx free of a ppxlib dep.

mlasson:
> You should be able to remove the ref mark_as_handled_manually. It was introduced when adding compatibility with ppxlib while leaving the ppx free of a ppxlib dep.

Thanks for the remark !

nojb:
Just to confirm, the `gen_js_api` binary can still be used to generate an implementation from an interface as before?

mlasson:
> Just to confirm, the gen_js_api binary can still be used to generate an implementation from an interface as before?

Yes !

rgrinberg:
This is awesome @mlasson. Is there a release pending?

mlasson:
> This is awesome @mlasson. Is there a release pending?

I'll trigger one next week.

PR #152, opened 2021-07-24
[@@js.invoke] attribute to call the global object as a function

cannorin:
We would want to bind to a global object with methods as a OCaml module with `js.scope`.
Sometimes, however, the global object is a "callable" one, which can also be called as a function.

A good example is the `Number` class/global object in ES5. `Number` is a global object which can be used as a constructor and can also be used as a function. Today we would have to bind to it as:

```ocaml
(* number.mli *)

type t = private Ojs.t

val toString: t -> ?radix:int -> unit -> float [@@js.call]
val toFixed: t -> ?fractionDigits:int -> unit -> float [@@js.call]
val toExponential: t -> ?fractionDigits:int -> unit -> float [@@js.call]
val toPrecision: t -> ?precision:int -> unit -> float [@@js.call]
val valueOf: t -> float [@@js.call]

module [@js.scope "Number"] Static : sig
  val create: 'any -> t [@@js.create]

  val min_value: float [@@js.global "MIN_VALUE"]
  val max_value: float [@@js.global "MAX_VALUE"]
  val nan: float [@@js.global "NaN"]
  val negative_infinity: float [@@js.global "NEGATIVE_INFINITY"]
  val positive_infinity: float [@@js.global "POSITIVE_INFINITY"]
end

val number: 'any -> float [@@js.global "Number"]
```

The problem is we can't `js.scope` the entire file because we lack the function counterpart of `js.create`, which calls the global object as a constructor.

This PR introduces a new attribute `js.invoke` to fill the gap:

|                                     | as a function          | as a constructor       |
|-------------------------------------|------------------------|------------------------|
| call the first argument             | `[@@js.apply]`         | `[@@js.apply_newable]` |
| call the global object              | `[@@js.invoke]` **(NEW)**        | `[@@js.create]`        |
| call a member of the first argument | `[@@js.call "name"]`   | N/A                    |
| call a member of the global object  | `[@@js.global "name"]` | `[@@js.new "name"]`    |

With this, we would be able to rewrite the above binding as:

```ocaml
[@@@js.scope "Number"]

type t = private Ojs.t

val toString: t -> ?radix:int -> unit -> float [@@js.call]
val toFixed: t -> ?fractionDigits:int -> unit -> float [@@js.call]
val toExponential: t -> ?fractionDigits:int -> unit -> float [@@js.call]
val toPrecision: t -> ?precision:int -> unit -> float [@@js.call]
val valueOf: t -> float [@@js.call]

val create: 'any -> t [@@js.create]
val number: 'any -> float [@@js.invoke]

val min_value: float [@@js.global "MIN_VALUE"]
val max_value: float [@@js.global "MAX_VALUE"]
val nan: float [@@js.global "NaN"]
val negative_infinity: float [@@js.global "NEGATIVE_INFINITY"]
val positive_infinity: float [@@js.global "POSITIVE_INFINITY"]
```


nojb:
cc @mlasson

VALUES.md
```diff
@@ -54,6 +54,30 @@ Supported forms
   val fooConstructor: FooConstructor.t [@@js.global "Foo"]
   ```
 
+  When the "callable" object you want to bind to is a global object, the `[@@js.invoke]`
+  attribute along with the `[@js.scope]` attribute (see below) may be used to call it.
+
+  For instance, you can write
+  ```ocaml
+    module[@js.scope "JavaScriptClassName"] C : sig
+      val invoke: T1 -> ... -> Tn [@@js.invoke]
```
mlasson: Minor suggestion:
```suggestion
      val invoke: T1 -> ... -> Tn -> t [@@js.invoke]
```
So there are N arguments :).

VALUES.md
```diff
@@ -54,6 +54,30 @@ Supported forms
...
+    end
+
+    (* usage *)
+    let x = C.invoke arg1 ... argn
+  ```
+  instead of
+  ```ocaml
+    module C : sig
+      type t
+      val apply: t -> T1 -> ... -> Tn [@@js.apply]
```
mlasson:
```suggestion
      val apply: t -> T1 -> ... -> Tn -> t [@@js.apply]
```

mlasson:
Thanks for the PR. It's great.

PR #153, opened 2021-07-30
Support recursive modules

### LINKED ISSUE #67: Would recursive modules be difficult (if possible) to implement?
I have something like this…

```ocaml  
module rec Contact : sig
   type t
   val get_contact_groups: t -> ContactGroup.t list [@@js.call]
end

and ContactGroup : sig
   type t
   val get_contacts: t -> Contact.t list [@@js.call]
end
```

I solved the problem by getting type Contact.t out of Contact module.
But I do not know if this solution will always be enough?

Best regards

cannorin:
Closes #67.

We would want to bind to JS classes and TS interfaces as OCaml modules.
However, classes and interfaces often have recursive references to each other, making it difficult to bind in OCaml.

```js
class Foo {
  constructor(name) { this.name = name; }
  describe() { return "Foo:".concat(this.name); }
  toBar() { return new Bar(this.name); }
}

class Bar {
  constructor(name) { this.name = name; }
  describe() { return "Bar:".concat(this.name); }
  toBar() { return new Foo(this.name); }
}
```

This PR adds support for [recursive modules](https://ocaml.org/manual/manual024.html) as input.
With this PR, we can easily bind to the above example as follows:

```ocaml
module [@js.scope "Foo"] rec Foo : sig
  type t = private Ojs.t
  val t_of_js: Ojs.t -> t
  val t_to_js: t -> Ojs.t
  val create: string -> t [@@js.create]
  val describe: t -> string [@@js.call "describe"]
  val to_bar: t -> Bar.t [@@js.call "toBar"]
end

and [@js.scope "Bar"] Bar : sig
  type t = private Ojs.t
  val t_of_js: Ojs.t -> t
  val t_to_js: t -> Ojs.t
  val create: string -> t [@@js.create]
  val describe: t -> string [@@js.call "describe"]
  val to_foo: t -> Foo.t [@@js.call "toFoo"]
end
```


cannorin:
I will do that myself once I have a time but feel free to fix the missing newline at the end of file, I thought it would be added automatically by my editor but it didn't work.

mlasson:
That's neat. Let's merge it !

PR #154, opened 2021-08-06
Upgrade to ocaml/setup-ocaml@v2

smorimoto:


mlasson:
Thank you very much !

PR #164, opened 2022-01-25
Switch to js_of_ocaml.4.0

hhugo:
see https://github.com/ocaml/opam-repository/pull/20533
and https://github.com/ocsigen/js_of_ocaml/pull/1224

### DIFF
--- a/lib/ojs_exn.ml
+++ b/lib/ojs_exn.ml
@@ -2,20 +2,15 @@
 (* See the attached LICENSE file.                                         *)
 (* Copyright 2015 by LexiFi.                                              *)
 
-type t = Ojs.t
+type t = Jsoo_runtime.Error.t
 
-let name x = Ojs.string_of_js (Ojs.get_prop_ascii x "name")
-let message x = Ojs.string_of_js (Ojs.get_prop_ascii x "message")
-let stack x = Ojs.option_of_js Ojs.string_of_js (Ojs.get_prop_ascii x "stack")
-let to_string x = Ojs.string_of_js (Ojs.call x "toString" [||])
+external coerce : t -> Ojs.t = "%identity"
+let name x = Ojs.string_of_js (Ojs.get_prop_ascii (coerce x) "name")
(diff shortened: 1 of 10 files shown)


lib/ojs_exn.ml
```diff
@@ -2,20 +2,15 @@
...
+let to_string x = Ojs.string_of_js (Ojs.call (coerce x) "toString" [||])
 
-exception Error of t
-
-let () = Callback.register_exception "jsError" (Error (Ojs.obj [||]))
-
-(* The js_of_ocaml runtime expects to have this registered.
-   So it's probably a bad idea to use both this Ojs_exn module
-   and the js_of_ocaml standard library. *)
+exception Error = Jsoo_runtime.Error.Exn
```
nojb, line 13: Is it the case that any code that used to catch `Ojs_exn.Error` will continue to work as before?

hhugo: Yes, as you can see, I haven't touched the interface. I'm just re-exporting the exception under a different name. The registration is done on the jsoo_runtime side and allow jsoo, gen_js_api and brr to coexist without interfering.

mlasson: We could also get rid of implements "externals" in the Ojs module with Jsoo_runtime... or maybe we could get rid of (or deprecate) the whole Ojs module and directly use Jsoo_runtime. But that may be another pull request. 

I'm totally ok to merge this one as soon as possible and make an opam release just to quickly have a version of gen_js_api compatible with js_of_ocaml 4 on opam.

nojb: @mlasson can you take care of merging this PR? Does anything need to be done for the CI to work?

mlasson: Hopefully, it will work when https://github.com/ocaml/opam-repository/pull/20533 will be merged (the CI does not know yet how to fetch js_of_ocaml 4.0 with out a "pinning it").

hhugo:
I've removed the -no-check-prim usage as it should no longer be necessary.

PR #165, opened 2022-01-31
Allow n-ary constructors in [@js.union] (treat them as tuple)

cannorin:
Sometimes we encounter a function one of which arguments is a union of TypeScript tuple and other types:
```typescript
export class ParameterInformation {
    ...
    constructor(label: string | [number, number], documentation?: string | MarkdownString);
}
```
(the above example is from https://github.com/microsoft/vscode/blob/1a57cb85407249f380f0ebfb34c748a960e5430a/src/vscode-dts/vscode.d.ts#L3889)

We would want to bind to such argument using `[@js.union]`, but the current implementation prohibits that:
```ocaml
module [@js.scope "ParameterInformation"] ParameterInformation : sig
    type t
    val create: label:([`String of string | `Tuple of (int * int)] [@js.union]) -> ?documentation:([`String of string | `Markdown of Markdown.t] [@js.union]) -> unit -> t [@@js.create]
    (* this fails with error Constructor_in_union *)
end
```

This PR relaxes the limitation and allow n-ary constructors in `[@js.union]` by treating them as if they were tuple types.

> Inline records in `[@js.union]` still remains prohibited, but I think we could relax this restriction too?


cannorin:
@mlasson I would appreciate if you can create a new release once this PR is merged, as it will enable `ts2ocaml` to erate much cleaner bindings with this feature 🙂

smorimoto:
ping @mlasson @alainfrisch

cannorin:
Thank you 🙏

PR #166, opened 2022-04-29
Support first class modules to treat type variables safely

cannorin:
This PR allows value bindings to take a limited form of first-class modules as arguments:

```ocaml
val console_log:
  (module[@js] Ojs.T with type t = 'a) -> 'a -> unit
  [@@js.global "console.log"]
```

Users can now pass first-class modules to safely convert the values of types with variables:

```ocaml
console_log (module Ojs.String) "hello, world!";;
console_log (module Ojs.Int) 42;;
```

The first-class modules must have the form of `(module[@js] Ojs.T with type t = 'a)`, and cannot be used outside of value bindings.

This PR also introduces several pre-defined modules for each built-in type to make life easier:
```ocaml
(* ojs.ml *)

module String = struct
  type t = string
  let t_to_js = string_to_js
  let t_of_js = string_of_js
end

...

module List (A: T) = struct
  type t = A.t list
  let t_to_js = list_to_js A.t_to_js
  let t_of_js = list_of_js A.t_of_js
end

...
```


cannorin:
It seems CI is broken since https://github.com/LexiFi/gen_js_api/commit/347a18c1321e6bac18627760911c97169511dadc.

I also fixed it, and now it's failing because `ppxlib` somehow generates code in a slightly different format in *nix than in Windows (it's not a version difference: both uses `0.26.0`).

Do you think this a bug of `ppxlib`...?

cannorin:
It's also failing in `issues.ml` and `types.ml`, which I didn't edit: https://github.com/LexiFi/gen_js_api/runs/6227945812?check_suite_focus=true#step:7:481

It seems there is some problem on printing locally abstract types?

cannorin:
I also upgraded ppxlib to 0.26 and disabled dune cache for Windows.

cannorin:
I will revert the CI fixes and make a separate PR to make it easy to review.

cannorin:
CI is broken again...?

mlasson:
> I also fixed it, and now it's failing because ppxlib somehow generates code in a slightly different format in *nix than in Windows (it's not a version difference: both uses 0.26.0).
> Do you think this a bug of ppxlib...?

Weird. I accepted the test on windows/cygwin/ocaml 4.14/ppxlib.0.26 and it has the same output as the CI. 
what is your setup to promote tests ?

mlasson:
The PR is nice and well documented, thanks !
Out of curiosity, are you using this somewhere ?

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -385,7 +392,15 @@ let no_attributes attributes =
     ) attributes;
   attributes = []
 
-let rec parse_arg ~variance ctx lab ~global_attrs ty =
+type type_context =
+  {
+    type_params: label list;
+    packages: (label * label) list; (** reversed order, (local_name, module_name) *)
```
mlasson: Why `label` instead of `string` ? 
I know they are synonyms but it seems that label are only used in the context of polymorhic variants and objects.

cannorin:
> what is your setup to promote tests ?

I'm using Debian on WSL1, which should be the same as in a normal Debian.

Also by explicitly upgrading to ppxlib 0.26, the issue somehow disappeared. So I think we can regard it as solved?

> Out of curiosity, are you using this somewhere ?

I'm using gen_js_api as the output of https://github.com/ocsigen/ts2ocaml 😄 

ts2ocaml would not be possible without gen_js_api. Thank you!

mlasson:
Thanks again for this PR. 
I made a new release containing this new feature. 
Hope this helps.

PR #167, opened 2022-04-30
Fix CI

cannorin:
* Disables `opam-pin`.
* Disables `dune-cache` on Windows.
* Upgrades ppxlib to 0.26.
* Updates the expected test results with ppxlib 0.26.


cannorin:
@mlasson could you review this and #166?

> The failing macos checks are because dune somehow got SIGKILL while running.
> I think it should go away once we re-run the checks.

Issue #160, opened 2021-10-27
Dots in js.scope


vouillon:
We used to be able to write:
```ocaml
module PlacesService: sig
  type t
  val new_places_service: Map.t -> t [@@js.new]
  ...
end
[@js.scope "google.maps.places"]
```
which was equivalent to using nested scopes `[@js.scope "places"][@js.scope "maps"][@js.scope "google"]`.

Since release 1.07, this no longer works since the following piece of code is generated:
```ocaml
Ojs.get_prop_ascii Ojs.global "google.maps.places"
```

Is this the expected behavior, or is this a bug?

nojb:
cc @mlasson

cannorin:
I created a PR which would allow us to write `[@js.scope ("google", "maps", "places")]`: #168

mlasson:
Fixed by #168

PR #168, opened 2022-05-26
Support tuple on @js.scope

cannorin:
Related: #160

This PR makes `@js.scope` accept a tuple payload, which can be used to write a nested scope.

* The example from #160 now can be written as:
  ```ocaml
  module PlacesService: sig
    type t
    val new_places_service: Map.t -> t [@@js.new]
    ...
  end
  [@js.scope ("google", "maps", "places")]
  ```
  
* Custom global object is also supported: assume `fs: require('fs')`, then we can now write
   ```ocaml
   [@@@js.scope (Imports.fs, "promises")]
   ```


alainfrisch:
This feature is not completely discoverable :-)  Is there a natural place in the doc where it could be mentioned?

mlasson:
Fair enough, I opened https://github.com/LexiFi/gen_js_api/pull/169

PR #170, opened 2022-12-08
Wrap functions supplied to Ojs.iter_properties

vouillon:
To support effect handlers, we are changing the calling convention of OCaml functions. See https://github.com/ocsigen/js_of_ocaml/pull/1340. Wrapping Ocaml functions is now mandatory to use them from JavaScript.

### DIFF
--- a/lib/ojs.ml
+++ b/lib/ojs.ml
@@ -117,7 +117,10 @@ external new_obj_arr: t -> t -> t = "caml_ojs_new_arr"
 
 let empty_obj () = new_obj (get_prop_ascii global "Object") [||]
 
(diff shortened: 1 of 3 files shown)


nojb:
Hello @vouillon, thansk for the contribution! I am wondering if some adaptations are not needed in the bindings generated by `gen_js_api` itself. Do you happen to have checked?

hhugo:
@vouillon, @nojb, it would be nice to get this released, I think users are starting to use gen_js_api and ocaml 5 with effects.  See https://discuss.ocaml.org/t/dealing-with-joo-tramp-joo-args/11630/2

nojb:
OPAM release with this patch submitted at https://github.com/ocaml/opam-repository/pull/23488

PR #172, opened 2023-06-30
Use js_of_ocaml runtime directly

mlasson:
# Description 

This is PR is still in a draft state.
It proposes to reduce the Ojs module to the strict minimum : the to and of functions. 
The ppx extension should be able to directly use the Jsoo_runtime.Js.module to handle pure JavaScript values.


mlasson:
The PR is responsible for CI failing on ocaml < 4.13.
See https://github.com/ocsigen/js_of_ocaml/pull/1485.

examples/test/main.ml
```diff
@@ -208,14 +216,23 @@ end = struct
...
+    let o = Jsoo_runtime.Js.obj [||] in
+    List.iter (fun (k, v) -> Jsoo_runtime.Js.set o (Ojs.string_to_js k) (ml2js v)) l;
     o
 
+  include
+    [%js:
+      val getOwnPropertyNames: Ojs.t -> string array [@@js.global "Object.getOwnPropertyNames"]
+    ]
+
+  let iter_properties obj f =
```
hhugo: code duplicated, already present above

lib/ojs.ml
```diff
@@ -13,7 +13,12 @@
 *)
 
 
-type t
+type t = Jsoo_runtime.Js.t
+
+external pure_js_expr: string -> t = "caml_pure_js_expr"
+let global = pure_js_expr "joo_global_object"
```
hhugo: We now use `globalThis` inside js_of_ocaml.

lib/ojs.ml
```diff
@@ -13,7 +13,12 @@
 *)
 
 
-type t
+type t = Jsoo_runtime.Js.t
+
+external pure_js_expr: string -> t = "caml_pure_js_expr"
```
hhugo: Why not use the external declared in jsoo_runtime ?

node-test/bindings/expected/arrays.ml
```diff
@@ -7,14 +7,20 @@ module JsArray(E:Ojs.T) =
     and t_to_js : t -> Ojs.t = fun (x1 : Ojs.t) -> x1
     let (create : unit -> t) =
       fun () ->
-        t_of_js (Ojs.new_obj (Ojs.get_prop_ascii Ojs.global "Array") [||])
+        t_of_js
+          (Jsoo_runtime.Js.new_obj
+             (Jsoo_runtime.Js.get
+                (Jsoo_runtime.Js.pure_js_expr "joo_global_object")
```
hhugo: globalThis

node-test/bindings/expected/arrays.ml
```diff
@@ -38,51 +45,65 @@ module JsArray2 =
     let (create : unit -> 'a t) =
       fun () ->
         t_of_js Obj.magic
-          (Ojs.new_obj (Ojs.get_prop_ascii Ojs.global "Array") [||])
+          (Jsoo_runtime.Js.new_obj
+             (Jsoo_runtime.Js.get
+                (Jsoo_runtime.Js.pure_js_expr "joo_global_object")
```
hhugo: globalThis

node-test/bindings/expected/arrays.ml
```diff
@@ -38,51 +45,65 @@ module JsArray2 =
...
+                (Obj.magic "Array")) [||])
     let (create' : (module Ojs.T with type t = 'a) -> 'a list -> 'a t) =
       fun (type a) ->
         fun ((module A)  : (module Ojs.T with type t = a)) ->
           fun (x12 : a list) ->
             t_of_js A.t_of_js
-              (Ojs.new_obj_arr (Ojs.get_prop_ascii Ojs.global "Array")
+              (Jsoo_runtime.Js.new_obj_arr
+                 (Jsoo_runtime.Js.get
+                    (Jsoo_runtime.Js.pure_js_expr "joo_global_object")
```
hhugo: globalThis

node-test/bindings/expected/arrays.ml
```diff
@@ -38,51 +45,65 @@ module JsArray2 =
...
+                    (Obj.magic "Array"))
                  (let x13 =
-                    Ojs.new_obj (Ojs.get_prop_ascii Ojs.global "Array") [||] in
+                    Jsoo_runtime.Js.new_obj
+                      (Jsoo_runtime.Js.get
+                         (Jsoo_runtime.Js.pure_js_expr "joo_global_object")
```
hhugo: globalThis

node-test/bindings/expected/buffer.ml
```diff
@@ -6,49 +6,62 @@ and t_to_js : t -> Ojs.t = fun (x1 : Ojs.t) -> x1
 let (alloc : int -> t) =
   fun (x3 : int) ->
     t_of_js
-      (Ojs.call (Ojs.get_prop_ascii Ojs.global "Buffer") "alloc"
-         [|(Ojs.int_to_js x3)|])
+      (Jsoo_runtime.Js.meth_call
+         (Jsoo_runtime.Js.get
+            (Jsoo_runtime.Js.pure_js_expr "joo_global_object")
+            (Obj.magic "Buffer")) "alloc" [|(Ojs.int_to_js x3)|])
```
hhugo: globalThis here and below

node-test/bindings/expected/console.ml
```diff
@@ -3,27 +3,35 @@
 let (log : 'a -> unit) =
   fun (x1 : 'a) ->
     ignore
-      (Ojs.call (Ojs.get_prop_ascii Ojs.global "console") "log"
-         [|(Obj.magic x1)|])
+      (Jsoo_runtime.Js.meth_call
+         (Jsoo_runtime.Js.get
+            (Jsoo_runtime.Js.pure_js_expr "joo_global_object")
```
hhugo: globalThis here and below

ppx-lib/gen_js_api_ppx.ml
```diff
@@ -875,38 +873,60 @@ let split sep s =
...
-    ojs "set_prop" [o; ojs "string_to_js" [str s]; v]
+let is_ascii s =
+  let exception Break in
+  try
+    String.iter (fun c -> if Char.code c > 127 then raise Break) s;
+    true
+  with Break -> false
+
+let get_prop o s =
+  jsoo "get" [o; if is_ascii s then magic (str s) else ojs "string_to_js" [str s]]
```
hhugo: Is the generated code different when removing the `is_ascii` special case ? I would expect the jsoo compiler to do the job already.

lib/ojs.ml
mlasson: Like @dbuenzli (See https://github.com/dbuenzli/brr/issues/45), I am not sure if exposing these functions as "external" in the interface would impact the performance or not.

lib/ojs.ml
mlasson: Do you think it could be a good idea to add this in "jsoo_runtime" ?
And what about null & undefined ?

lib/ojs.ml
hhugo: I would say it's better use externals because it allows jsoo to perform more optimisations.

Note that jsoo_runtime doesn't have an mli / it export external in its API.
I think it doesn't change any optimisation if you use the external defined locally or in another file/module

lib/ojs.ml
hhugo: I don't know. maybe @dbuenzli has an opinion.

lib/ojs.ml
nojb: My understanding is as follows:

1) when using the bytecode compiler `ocamlc`, if `ojs` does _not_ expose the `external` declaration, clients of this module will indeed pay a (small) penalty, roughly one function call + one memory indirection.
2) however, JSOO is clever enough to inline the external declaration across modules, which removes the penalty when targeting Javascript.

Conclusion: you can reference the externals declared in `jsoo_runtime` without any loss of efficiency.

lib/ojs.ml
dbuenzli: No opinion :-) FWIW I don't expose `pure_js_expr` and here are its usages in `brr`.

```
> git grep pure_js_expr
src/jstr.ml:external jv_pure_js_expr : string -> 'a = "caml_pure_js_expr"
src/jstr.ml:let jv_global = jv_pure_js_expr "globalThis"
src/jstr.ml:let iterator : jv = jv_pure_js_expr "Symbol.iterator"
src/jv.ml:external pure_js_expr : string -> 'a = "caml_pure_js_expr"
src/jv.ml:let null = pure_js_expr "null"
src/jv.ml:let undefined = pure_js_expr "undefined"
src/jv.ml:let global = pure_js_expr "globalThis"
src/jv.ml:let true' = pure_js_expr "true"
src/jv.ml:let false' = pure_js_expr "false"
src/jv.ml:  let symbol : jv = pure_js_expr "Symbol.iterator"
```

It seems at the moment I have no equality in `Jv`'s implementation between `Jv.t` and `Jsoo_runtime.Js.t` so I'd need this to be able to use such values from `Jsoo_runtime` but I don't think it would pose any problem.

ppx-lib/gen_js_api_ppx.ml
mlasson: You're right, the compiler seems to take care of that. 

```ocaml 
open Jsoo_runtime

let global : Js.t =
  Js.pure_js_expr "globalThis"

external get_external: Js.t -> string -> Js.t = "caml_js_get"

let () =
  ignore (get_external global "console");
  ignore (Js.get global (Obj.magic "console"));
  ignore (Js.get global (Js.string "console"))
```
compiles as
```javascript
(function
  (globalThis){
   "use strict";
   var
    runtime = globalThis.jsoo_runtime,
    cst_console = "console",
    global = globalThis;
   global[cst_console];
   global[cst_console];
   global.console;
   var Dune_exe_Test = [0, global];
   runtime.caml_register_global(4, Dune_exe_Test, "Dune__exe__Test");
   return;
  }
  (globalThis));
```

mlasson: It's a bit surprising that they are not compiled exactly in the same way actually. 
(also note that is you replace "console" by "con sole" then they are all compiled as `global[cst_console]` which is not surprising). 
It seems that the first strategy as better potential for "minimization". But maybe it is slower to interpret ? I don't know.

lib/ojs.ml
mlasson: > It seems at the moment I have no equality in Jv's implementation between Jv.t and Jsoo_runtime.Js.t

If this PR gets merged, I think that adding this equality in jv would improve a bit the "interoperability" between brr and gen_js_api !

dbuenzli: Well I won't add an equality at the `.mli` level, this makes error messages from the compiler more confusing, verbose and harder to understand for end users for little return most of the time.

I'd only add the equality at the `.ml` level so that I can reexport the values for `null`, `undefined` if they are defined in `Jsoo_runtime.Js.t`. But I can provide `{to,of} convertors in `Jv` if that's desirable.

mlasson: Ok ! That makes sense. Thank you for your answer.

node-test/bindings/container.ml
```diff
@@ -2,13 +2,21 @@ module StringMap = struct
...
   let t_to_js ml2js l =
-    let o = Ojs.empty_obj () in
-    iter (fun k v -> Ojs.set_prop o (Ojs.string_to_js k) (ml2js v)) l;
+    let o = Jsoo_runtime.Js.obj [||] in
+    iter (fun k v -> Jsoo_runtime.Js.set o (Ojs.string_to_js k) (ml2js v)) l;
     o
 
+  include
+    [%js:
+      val getOwnPropertyNames: Ojs.t -> string array [@@js.global "Object.getOwnPropertyNames"]
```
hhugo, line 11: Do you think this should be part of jsoo_runtime ?

lib/ojs.ml
alainfrisch: > however, JSOO is clever enough to inline the external declaration across modules, which removes the penalty when targeting Javascript.

This would not apply, I guess, if we use jsoo's separate compilation.  @nojb : is that also your understanding?

nojb: Haven't tested it, but that would be my guess indeed.

hhugo: This is not my understanding. As long as ocamlc is aware that a symbol is an external, it will be reflected in the bytecode and js_of_ocaml will be able to perform optimisations based on primitive names. It doesn't change anything if the external is declared locally or in a separate compilation unit.

hhugo: the following
```ocaml
let a = Jsoo_runtime.Js.pure_js_expr "thisIsExported"`
```
with separate compilation, gets translated into 
```js
// Generated by js_of_ocaml
//# buildInfo:effects=false, kind=cmo, use-js-string=true, version=5.3.0+git-5.3.0-18-gdb77ca4f04

//# unitInfo: Provides: B
//# unitInfo: Requires: Jsoo_runtime
(function
  (globalThis){
   "use strict";
   var runtime = globalThis.jsoo_runtime, a = thisIsExported, B = [0, a];
   runtime.caml_register_global(1, B, "B");
   return;
  }
  (globalThis));
```

node-test/bindings/container.ml
mlasson: No, I think this counts as a "binding" to a global function, I don't think such things should be in jsoo_runtime. 

Here's a small list of things I think could go there though: 
   - The global object.
   - Conversion between Error.t and Js.t ? Or making it transparently equal to Js.t ?
   - The "int -> Js.t" and "Js.t -> int" conversion functions ?

lib/ojs.ml
mlasson: - `get.ml`
```ocaml
open Jsoo_runtime
let get = Js.get
external get2 : Js.t -> Js.t -> Js.t = "caml_js_get"
```
- `get.mli`
```ocaml
open Jsoo_runtime
val get: Js.t -> Js.t -> Js.t
external get2 : Js.t -> Js.t -> Js.t = "caml_js_get"
```
is compiled as: 
```javascript
(function
  (globalThis){
   "use strict";
   var
    runtime = globalThis.jsoo_runtime,
    get = runtime.caml_js_get,
    Dune_exe_Get = [0, get];
   runtime.caml_register_global(0, Dune_exe_Get, "Dune__exe__Get");
   return;
  }
  (globalThis));
```
See how get2 is not even present in the "javascript module" Dune_exe_Get. 

Then if a client of this module: 
```ocaml
open Jsoo_runtime
let global = Js.pure_js_expr "globalThis"
let () =
  ignore (Get.get global (Js.string "console"));
  ignore (Get.get2 global (Js.string "console"))
```
is compiled as: 
```javascript
// Generated by js_of_ocaml
//# buildInfo:effects=false, kind=cmo, use-js-string=true, version=5.3.0+git-5.3.0-16-g32a8f920-dirty

//# unitInfo: Provides: Dune__exe__Test
//# unitInfo: Requires: Dune__exe__Get, Jsoo_runtime
(function
  (globalThis){
   "use strict";
   var runtime = globalThis.jsoo_runtime, cst_console = "console";
   function caml_call2(f, a0, a1){
    return (f.l >= 0 ? f.l : f.l = f.length) == 2
            ? f(a0, a1)
            : runtime.caml_call_gen(f, [a0, a1]);
   }
   var
    global_data = runtime.caml_get_global_data(),
    Dune_exe_Get = global_data.Dune__exe__Get,
    global = globalThis;
   caml_call2(Dune_exe_Get[1], global, cst_console);
   global.console;
   var Dune_exe_Test = [0, global];
   runtime.caml_register_global(4, Dune_exe_Test, "Dune__exe__Test");
   return;
  }
  (globalThis));
```

If I am not wrong, it means that using "Jsoo_runtime" instead of "externals" to implement Jv in https://github.com/dbuenzli/brr/issues/45 would indeed have a negative impact on the performance in "dev mode", no ?

hhugo: externals exported in the mli of `Jv` need to stay `external` so that when referencing `Jv` from another unit, ocamlc (and jsoo) know about these externals.

mlasson:
Thanks for all the discussions ! 
I will close this PR, I no longer think this is a good idea !

hhugo:
Can you maybe elaborate ? Why isn't it a good idea?

mlasson:
Apologies for being a bit abrupt, and thank you for asking for more details.

There were a few things I aimed to accomplish with this PR:
1. I wanted to minimize the number of things that are added to "ojs." My intention was to avoid turning Ojs into just another small "Javascript Value Library". I wanted to prevent further fragmentation of the ecosystem.
2. I aimed to make the Ojs.t type transparent to improve interoperability.
3. I wanted to use the "Jsoo_runtime" module to avoid direct dependencies on primitives.

I'm not entirely convinced that "2" is still a good idea. No one specifically requested it, and by making it transparent, it could become more difficult to change it in the future. Perhaps, we will want to use the type of values from a library that the community will eventually adopt.

If we keep the type as Opaque, we'll need to keep the various primitives available to users and code generator for handling it. So it's not really possible to minimize it (unless we add "to_runtime"/"of_runtime" coercion functions everywhere in the code generator ?).

Lastly, the fear of impacting the performance in dev mode scared me away from replacing all externals.

hhugo:
Ok, thanks for the details

PR #173, opened 2023-07-25
Compatibility with Wasm_of_ocaml

vouillon:
[Wasm_of_ocaml](https://github.com/ocaml-wasm/wasm_of_ocaml) uses a different representation for OCaml floats and JavaScript numbers, so we need explicit coercions (which are still optimized away by Js_of_ocaml).

Also, since JavaScript values get boxed, physical equality can no longer be used to test whether a value is `undefined`.
But as `undefined == null`, I have just simplified the test in `option_of_js`.

The WebAssembly implementation of the runtime functions `caml_ojs_wrap_fun_arguments` and `caml_ojs_iterate_properties` is provided by `wasm_of_ocaml`.

### DIFF
--- a/lib/ojs.ml
+++ b/lib/ojs.ml
@@ -91,7 +91,7 @@ let list_to_js f l =
   array_to_js f (Array.of_list l)
 
 let option_of_js f x =
-  if equals x null || x == undefined then None
+  if equals x null then None
   else Some (f x)
 
 let option_to_js f = function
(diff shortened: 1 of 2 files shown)


mlasson:
Thanks for the PR ! wasm_of_ocaml looks super cool !

OlivierNicole:
Hello, do you intend to make a new release with these changes? Js_of_ocaml will probably release a breaking change regarding float conversions soon-ish.

OlivierNicole:
I see that it’s done, that was fast! Thanks a lot!

PR #174, opened 2024-06-01
Add `[@js.dict]`, `[@@@js.require]`, `[@@js.capitalize]`

nojb:
This PR adds the possibility of annotating an OCaml type expression `(string * ty) list` with `[@js.dict]` to indicate that values of this type should be mapped to JS objects in the obvious way (and vice-versa).

### DIFF
--- a/TYPES.md
+++ b/TYPES.md
@@ -16,6 +16,9 @@ The following types are supported out-of-the-box:
  - Sequences of JS-able types: `array` and `list`, both mapped to JS
    arrays (which are assumed to be indexed by integers 0..length-1).
 
+ - Dictionaries of JS-able types: `(string * 'a) list` mapped to
+   a JS object.
+
  - Options on JS-able types.  They are mapped to the same type as
    their parameter: `None` is mapped to JS `null` value, and both
    `null` and `undefined` are mapped back to `None`.  This encoding
@@ -204,6 +207,20 @@ implementation).  Mutually recursive type declarations are supported.
 - Sum type declaration, mapped to enums (see Enums section).
 
 - Sum type declaration with non constant constructors, mapped to records with a discriminator field (see Sum types section).
+
+
+- Association lists, mapped to JS objects
+
+  It is possible to annotate an OCaml type declaration of the form
+  ```
+  (string * ty) list
+  ```
+  (where `ty` is any JS-able type) with `[@js.dict]`. When this is done, values
+  of this type will be mapped to JS objects in the obvious way.
+
+  ```ocaml
+  type t = { headers: ((string * string) list [@js.dict]) }
+  ```
 
 - Arbitrary type with custom mappings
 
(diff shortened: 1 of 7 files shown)


nojb:
The second commit adds support for `[@@@js.require "foo"]`, which generates in the implementation side a let-binding
```
let ___require = Ojs.apply (Ojs.variable "require") [|Ojs.string_to_js "foo"|]
```
__and__ sets the global object to be `__require`. This simplifies binding Node libraries, eg
```
[@@@js.require "crypto"]
type hash
val create_hash: string -> hash [@@js.global]
```
would bind https://nodejs.org/api/crypto.html#cryptocreatehashalgorithm-options.

The generated javascript contains `require("crypto")` (both in the `dev` and `release` profiles) with a literal string argument (which seems important, see https://github.com/LexiFi/gen_js_api/issues/66#issuecomment-526844888, cc @jchavarri), but only if the string `"crypto"` does not appear elsewhere in the file (because JSOO shares constants whenever it can), opened https://github.com/ocsigen/js_of_ocaml/issues/1619 about this.

nojb:
The third (and last!) commit adds the possibility of telling JSOO to use `CamelCase` instead of `camelCase` for the generated JS names, which is useful to bind certain APIs and much cleaner than using `[@js "CamelCase"]` everywhere. (The same could be achieved by naming the OCaml labels `_camel_case`, but it is not very nice...).

nojb:
Prior art for the `require` part: #66 and #68

nojb:
Friendly ping.

mlasson:
### `@js.dict`

I am not sure we want to add the `[@js.dict]` mainly because I am not super fond of the "Ojs.iter_properties" (I would love to deprecate it actually). 

To implement your dictionnaries, you should be able to do what you want by creating a module like: 
```
module Dict = struct 
  type 'a t = (string * 'a) list 
  let t_of_js f t =
    let l = ref [] in
    iter_properties t (fun k -> l := (k, f (get_prop_ascii t k)) :: !l);
    !l

  let t_to_js f x =
    let t = empty_obj () in
    List.iter (fun (k, v) -> set_prop_ascii t k (f v)) x;
    t
end
```
Then you can replace  `((string * t) list [@js.dict]` by `t Dict.t`. 
I've just realized that this example was mentioned in "LOW_LEVEL_BINDING.md . 

### `@js.require`

The current way to live without require is to add a custom JavaScript stub that enrich the global object with some object imported from modules (using one of the various module system found in the JavaScript ecosystem). I am not completely against providing an out-of-the box support a require annotation but it will be better to check if it works not only with node but also with bundling tools like [browserify ](https://browserify.org/) and [webpack](https://webpack.js.org/).

### `@js.capitialize`

Do you think it would be interesting to support a module annotation (ie. @@@) for that ?

nojb:
Thanks for the review! The changes here are not critical; I suggest we punt this and revisit it later if/as needed.

PR #175, opened 2024-10-13
Support for simple module construction.

sbriais:
This PR adds support for

- module aliases `module A = B`

This can be convenient when writing interfaces and referring to other components.

For instance, in another mli

```
module Event = Js_browser.Event

val foo: Event.t -> string -> unit [@@js.call]
```

This way, we can just write `Event` and not `Js_browser.Event` everywhere.


mlasson (approved):
i added a small piece of documentation, feel to free to merge or amend it !

PR #176, opened 2024-11-19
remove references to joo_global_object

hhugo:
It is deprecated. The js_of_ocaml-compiler will emit a warning when it sees `joo_global_object` in its next release.


nojb:
Thanks for the PR (and sorry for the lag). We will review shortly.

PR #179, opened 2024-12-27
Update GitHub Actions

smorimoto:
- Automate GitHub Actions updates with Dependabot
- Update all GitHub Actions
- Enable testing on macOS and Windows
  - With the benefit of `setup-ocaml` v3!
- Add `lint-opam` to make sure there are no errors in the opam file or dune configuration
- Remove `conf-npm` from depends
  - As long as we use `setup-node` in CI, we don't need it, and the system package manager always installs obsolete Node.js...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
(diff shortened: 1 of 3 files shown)


mlasson (approved):
Thanks this is neat !

PR #180, opened 2025-07-01
Fix for OCaml 5.3

hhugo:
This is due to the following change in OCaml 5.3
```
- #11129, #11148: enforce that ppxs do not produce `parsetree`s with
  an empty list of universally quantified type variables
  (`. int -> int` instead of `'a . int -> int'`)
  (Florian Angeletti, report by Simmo Saan, review by Gabriel Scherer)
```

I currently get  the following error.
```
File "_none_", line 1:
Error: broken invariant in parsetree: Explicit universal type quantification cannot be empty.
```


mlasson:
I don't mind much but I don't really understand the diff on the pretty printed trees. 
Do you know if this is something known ?

nojb:
> I currently get the following error.

Do you have an example that triggers this error? We use 5.3 internally at LexiFi and have not seen this error.

hhugo:
The testsuite in this repo triggers the error.
```
File "node-test/bindings/promise.mli", line 20, characters 4-40:
20 |     type wrap = { content: Ojs.t }[@@js]
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: broken invariant in parsetree: Explicit universal type quantification cannot be empty.
```

Maybe it only triggers with ppxlib.0.36 ?

nojb:
I confirm that it does not trigger with 0.35.

> Maybe it only triggers with ppxlib.0.36 ?

I think this is likely the case, but I can't confirm this at the moment (a bunch of packages are incompatible with ppxlib 0.36.0).

hhugo:
> I don't mind much but I don't really understand the diff on the pretty printed trees. Do you know if this is something known ?

The diff is unrelated to the PR. I suspect you just haven't run the CI  with OCaml 5.3 yet.

hhugo:
> > I don't mind much but I don't really understand the diff on the pretty printed trees. Do you know if this is something known ?
> 
> The diff is unrelated to the PR. I suspect you just haven't run the CI with OCaml 5.3 yet.

In fact, the diff is due to ppxlib.0.36 that bumps its internal ast to OCaml ~~5.3~~ 5.2.

hhugo:
I think I have more explanations.
Ppxlib.0.36 bumped its internal ast from 5.0 to 5.2. (not 5.3 as I previously stated)

When using OCaml 5.3 and ppxlib 0.35, ppxlib would apply to following conversion 5.3 -> 5.0 -> ppx -> 5.0 -> 5.3.
When converting 5.0 to 5.3, it would normalize (or fix) the ast containing `Ptyp_poly ([], ..)` .

I believe this is done in `migrate_500_501.ml` with 
```ocaml
    let value_pattern =
      match p with
      | {
       Ast_500.Parsetree.ppat_desc =
         Ppat_constraint
           ( ({ ppat_desc = Ppat_var _ } as pat),
             ({ ptyp_desc = Ptyp_poly (args_tyvars, rt) } as ty_ext) );
       ppat_attributes = [];
      } ->
          assert (match rt.ptyp_desc with Ptyp_poly _ -> false | _ -> true);
          let ty = match args_tyvars with [] -> rt | _ -> ty_ext in
          `Var (pat, args_tyvars, rt, ty)
      | {
       Ast_500.Parsetree.ppat_desc = Ppat_constraint (pat, rt);
       ppat_attributes = [];
      } ->
          `NonVar (pat, rt)
      | _ -> `None
    in
```
  
When using ppxlib.0.36, we perform the following conversion: 5.3 -> 5.2 -> ppx -> 5.2 -> 5.3. and do not go through a code pass that would "fix" the ast.

hhugo:
https://github.com/ocaml/opam-repository/pull/28121

PR #181, opened 2025-07-30
Use ppxlib version 0.37 at least

hhugo:
See https://github.com/ocaml/opam-repository/pull/28160

One issue is that ppxlib 0.36 is currently broken an drops pvb_constraint on the floor when printing the AST.
So maybe we should bump the constraint to use the unreleased 0.36.1 version.

cc @NathanReb 

EDIT: we now use ppxlib 0.37


nojb:
Thanks @hhugo! I believe ppxlib 0.37.0 will be released shortly. As soon as that happens, we can update the lower bound here and then merge.

nojb:
ppxlib 0.37 has been released. @hhugo: do you mind updating the version constraint here and then we can merge? Thanks much!

hhugo:
> ppxlib 0.37 has been released. @hhugo: do you mind updating the version constraint here and then we can merge? Thanks much!

Updated to ppxlib 0.37

nojb:
Merged, thanks! I will try to get a release out soon.

nojb:
> Merged, thanks! I will try to get a release out soon.

https://github.com/LexiFi/gen_js_api/releases/tag/v1.1.6
https://github.com/ocaml/opam-repository/pull/28721

